PR #26, opened 2021-11-17
Aggressive inlining

troelsfr:


qir/qat/Apps/Qat/Qat.cpp
```diff
@@ -59,6 +59,7 @@
 
 #include <dlfcn.h>
 
+#include <climits>
```
swernli, line 62: Now that `cfg.inlineParameter()` is used below, this include is no longer needed.

swernli (approved):
Changes look good to me!

qir/qat/Apps/Qat/Qat.cpp
troelsfr: Removed.

PR #39, opened 2021-11-24
Adding C++ stdlib

troelsfr:
This PR introduces a C++ implementation of the stdlib as an alternative to the C version.


qir/stdlib/src/stdlib.c
```diff
@@ -2,7 +2,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-
+/*
```
swernli, line 5: It looks like some of these changes may have accidentally made it into #40

qir/stdlib/src/stdlib.c
```diff
@@ -96,7 +97,37 @@ int8_t *__quantum__rt__array_create_1d(int32_t size, int64_t n)
...
+void __quantum__rt__array_update_reference_count(struct Array *arr, int32_t n)
+{
+  int64_t *r = (int64_t *)((int8_t *)arr + 2 * sizeof(int64_t));
+  *r += (int64_t)(n);
+  /*
+  if (*r <= 0)
+  {
+    free(arr);
+  }
+  */
```
swernli, line 130: With this commented out, does it mean arrays will never be freed?

qir/stdlib/src/stdlib.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  void __quantum__rt__array_update_alias_count(Array *arr, int32_t n)
+  {
+    arr->alias_count += n;
+  }
+
+  void __quantum__rt__array_update_reference_count(Array *arr, int32_t n)
+  {
+    arr->ref_count += n;
```
swernli, line 81: Likewise here. It seems nothing is stopping `ref_count` from going to negative numbers and/or an array with zero reference count from being leaked.

swernli (commented):
I think the handling for deleting an array was accidentally removed, and some of the changes to the original C stdlib got mixed in with the other active PR.

qir/stdlib/src/stdlib.c
troelsfr: This was a test to see if I could get things to perform better - resolving as it is not relevant for the latest version.

qir/stdlib/src/stdlib.cpp
troelsfr: The stdlib assumes that you write correct code. If we add too much logic it just becomes harder for LLVM to optimise things away.

qir/stdlib/src/stdlib.c
troelsfr: Will have a look.

swernli (approved):
Approved with a couple additional comments. I was able to try this new version of the stdlib locally and can confirm it works for me!

qir/qat/RemoveDisallowedAttributesPass/RemoveDisallowedAttributesPass.hpp
```diff
@@ -25,10 +25,20 @@ namespace quantum
 
         llvm::PreservedAnalyses run(llvm::Module& module, llvm::ModuleAnalysisManager& /*mam*/)
         {
+            std::string const LLVM_FUNC_NAME = "@llvm.";
             for (auto& fnc : module)
             {
                 std::unordered_set<String> to_keep;
+                auto                       name = static_cast<std::string>(fnc.getName());
 
+                // Skipping any LLVM function
```
idavis, line 34: Why are we leaving in commented out code?

qir/stdlib/lib/stdlib-cxx.ll
```diff
@@ -0,0 +1,621 @@
+; ModuleID = 'src/stdlib.cpp'
+source_filename = "src/stdlib.cpp"
```
idavis, line 2: I think we should be checking in the `bc` files instead of `ll`. /CC @swernli @bettinaheim

qir/stdlib/src/stdlib.cpp
```diff
@@ -0,0 +1,137 @@
...
+    auto &a   = *reinterpret_cast<int64_t *>(ret + 2 * sizeof(int64_t));
+    s         = n;
+    r         = 1;
+    a         = 0;
+    return ret + 3 * sizeof(int64_t);
+  }
+
+  void __quantum__rt__tuple_update_reference_count(int8_t *tuple, int32_t n)
+  {
+    auto  ptr = tuple - 3 * sizeof(int64_t);
```
idavis, line 51: What is `3`? Perhaps a named constant to clear it up?

qir/stdlib/lib/stdlib-cxx.ll
swernli: I'm torn on this one, because I do want to use IR as the source of truth if possible, but that means we end up checking in (in either format) the `target datalayout` and `target triple` of whoever last generated the file. I can't use this file directly in Linux because the target triple/datalayout is for Mac, so I have to delete those lines or rerun the compilation. Compiling again not only updates the file with my local datalayout/triple but also makes a few other minor changes to the rest of the file (differences in C++ -> IR with AppleClang vs Clang-11). I'm worried checking in IR in any form other than hand-written is going to cause headaches in git and confusion in expected usage.

idavis: @swernli  IIRC you have to manually set a datalayout during `llvm-as` to force a layout that isn't in the IR.

swernli: If a file doesn't have any data layout string then it just uses the LLVM default. From the [docs](https://llvm.org/docs/LangRef.html#langref-datalayout):

> When constructing the data layout for a given target, LLVM starts with a default set of specifications which are then (possibly) overridden by the specifications in the datalayout keyword.

But what I've observed is that since QAT combines the stdlib IR and the program IR, LLVM C-API will reject that if the two files have different datalayout or target triple, resulting in a runtime failure in QAT.

qir/stdlib/src/stdlib.c
troelsfr: Removed from #40.

qir/stdlib/Makefile
```diff
@@ -1,22 +1,47 @@
+lib/stdlib-cxx.ll:
```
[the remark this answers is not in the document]
troelsfr: From what I understand this is controlled from another repo. @bettinaheim is adding whitenoise. Could we restrict it to`.md` files to start with? For code files you need something slightly different as words typically are combined into a single token such a `a_class` or `awesomeFunction` (both correct, but neither will pass CSpell).

qir/stdlib/include/qir-stdlib.h
```diff
@@ -0,0 +1,23 @@
+#ifndef QIR_STDLIB_H
```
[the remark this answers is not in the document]
troelsfr: Yes, it does. I have not enabled CI for the stdlib yet as it was just a prototype last Wednesday, but will clean this up today.

qir/stdlib/include/qir-stdlib.h
```diff
@@ -0,0 +1,23 @@
...
+#define QIR_STDLIB_H
```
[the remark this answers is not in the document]
troelsfr: I mostly chose to stick with `#ifdef` as this is a C header and was not sure to which extend #pragmas has support. I will look into this and revise accordingly.

qir/qat/RemoveDisallowedAttributesPass/RemoveDisallowedAttributesPass.hpp
troelsfr: This is a mistake. This PR should not touch anything in `qat`.

qir/stdlib/src/stdlib.cpp
troelsfr: Yes, the entire file needs to be cleaned up.

qir/stdlib/tests/csuite/test.c
```diff
@@ -1,7 +1,5 @@
-void __quatum__hello_world();
 
 int main()
 {
-  __quatum__hello_world();
   return 0;
```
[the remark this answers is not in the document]
troelsfr: It is just a placeholder. I can add a comment saying that it is so, but I did not have time to write C tests last week and I am wondering whether we should just stick with C++ tests. They are easier to write and read. @swernli  and @bettinaheim  thoughts?

qir/stdlib/lib/stdlib-cxx.ll
troelsfr: @idavis @swernli  and @bettinaheim  the BC should be in the release artefacts IMO. The way I view it is that `ll` is equivalent to the "source" of the IR and bytecode is equivalent to the "binary" (even if there is a one-to-one correspondence). 

As such, I think it is questionable as to whether any IR should be checked in at all as they will always have platform dependence in form of the data layout and target machine. For instance, the IR will contain optimisation based on endianess assumptions that may not be true. Hence, the "right" thing to do would be to recompile the stdlib for the target platform (as we do for any other programming language).

troelsfr: Regardless, this is a broader discussion and out of scope for this PR. We possibly need to understand what makes an IR target agnostic and how link-time optimisation works before making a decision on this. For this reason, I suggest that we create a ticket for this discussion. Are you OK with that  @bettinaheim?

PR #40, opened 2021-11-24
QIS grouping

troelsfr:
This PR introduces a pass that groups `qis` functions.


qir/qat/GroupingPass/GroupingAnalysisPass.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+        static String const QIS_START;
+        static String const READ_INSTR_START;
+
+        // Construction and destruction configuration.
+        //
+
+        explicit GroupingAnalysisPass(GroupingPassConfiguration const& cfg)
+          : config_{cfg}
+        //    , logger_{logger}
```
swernli, line 46: Could this be removed or is it left as a hint for the future?

qir/qat/GroupingPass/GroupingPass.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    llvm::PreservedAnalyses GroupingPass::run(llvm::Module& module, llvm::ModuleAnalysisManager& mam)
+    {
+        auto& result = mam.getResult<GroupingAnalysisPass>(module);
+
+        for (auto* block : result.qc_cc_blocks)
+        {
+            expandBlockQuantumClassical(module, block);
+        }
```
swernli, line 251: I'm a little confused as to why this is needed in addition to the use of `expandBlockQuantumMeasureClassical`. Wouldn't any block that has quantum and classical but no measurement also be handled by `expandBlockQuantumMeasureClassical`? Using the before/after graph you created in the header file, I would've assumed that passing a block with no measurements would just result in empty "bb mc" and "bb tail cc" blocks, which collapses trivially to the specialized no measurements case. I'm unclear on what value processing blocks without measurements separately provides.

qir/stdlib/src/stdlib.c
```diff
@@ -2,7 +2,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-
+/*
```
swernli, line 5: It looks like these updates to the stdlib may have snuck in here inadvertently. Are these related to the grouping pass? If not, it would be better to not include them here and leave those for a separate PR.

swernli (commented):
This looks good so far! I had a few questions below, the one major one being a clarification on why having two versions of the grouping pass (one for blocks without any measurement, one for blocks with measurement) is needed.

In addition, it would be good to have this new feature disabled by default and allow folks to opt-in via flag until a generalized approach is ready. If that is already achievable in this PR then the flag may not be needed, but it could allow this merge sooner if the rest of the work to generalize the pass ends up being pretty big.

qir/qat/GroupingPass/GroupingPass.cpp
troelsfr: Yes, they can indeed. The two functions just reflect progressive work and the initial thought that they need to be handled separately. One thing worth noting though is that there may be merits to separating it into two functions as the former amounts to fire-and-forget for a async function where as the latter is fire and await. Whether this is something we should have support for is a question that we probably should consider.

qir/stdlib/src/stdlib.c
troelsfr: Yes, this should indeed just be the QIR from main.

qir/qat/Apps/Qat/Qat.cpp
```diff
@@ -235,6 +238,15 @@ int main(int argc, char** argv)
                 }
             });
 
+        generator->registerProfileComponent<GroupingPassConfiguration>(
+            "grouping", [](GroupingPassConfiguration const& cfg, ProfileGenerator* ptr, Profile& profile) {
+                auto& mam = profile.moduleAnalysisManager();
+                mam.registerPass([&] { return GroupingAnalysisPass(cfg); });
+                auto& ret = ptr->modulePassManager();
+                // ret.addPass(GroupingAnalysisPassPrinter());
```
swernli, line 246: nitpick: It would be good to include a comment as to why this line is left here, so folks know they can uncomment it if they want some additional printed output.

swernli (commented):
I'm not sure if it's something I'm doing wrong in invoking it, but I'm seeing behavior where the calls to `__quantum__qis__read_result__body` are not getting pulled down into the readout block as I'd expect them to be, but instead are left interleaved with the instructions that prepare the qubit and measure it. Is this intentional?

swernli (approved):
I've been able to verify the updated behavior across a few different examples, and it all look like it is working well!

PR #41, opened 2021-11-26
QIR Alliance PR check workflow

bettinaheim:
Workflow directly from the org template without modifications. The markdown linting rules are hence defined by the org.

### DIFF
--- a/.github/workflows/link_and_spell_check.yml
+++ b/.github/workflows/link_and_spell_check.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     name: "Check spelling, linting and links"
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
     - run: |
         wget https://github.com/qir-alliance/.github/archive/refs/heads/main.zip -O clone.zip &&
         unzip clone.zip && mv .github-main out && rm clone.zip &&
@@ -36,8 +36,6 @@ jobs:
       with:
         use-verbose-mode: 'yes'
       if: ${{ success() || failure() }}
-    - name: "Checkout"
-      uses: actions/checkout@v2
     - name: "Check spelling in changed files"
       # cspell:words streetsidesoftware
       uses: streetsidesoftware/cspell-action@v1.2.4
--- a/.github/workflows/link_and_spell_check.yml
+++ b/.github/workflows/link_and_spell_check.yml
@@ -1,0 +1,50 @@
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+    - main
+
+name: "Check spelling, linting and links"
+
+jobs:
+  check-markdown:
+    runs-on: ubuntu-latest
+    name: "Check spelling, linting and links"
+    steps:
+    - uses: actions/checkout@master
+    - run: |
+        wget https://github.com/qir-alliance/.github/archive/refs/heads/main.zip -O clone.zip &&
+        unzip clone.zip && mv .github-main out && rm clone.zip &&
+        mv out/workflow-templates/utils/ utils && rm -r out
+      shell: bash
+    - run: |
+        Get-ChildItem (Join-Path "utils" "problem-matchers") `
+          | ForEach-Object {
+            Write-Host "Adding matcher $_...";
+            Write-Host "::add-matcher::$_";
+          }
+      shell: pwsh
+    - name: markdownlint-cli
+      # cspell:words nosborn
+      uses: nosborn/github-action-markdown-cli@v2.0.0
+      with:
+        files: .
+        config_file: 'utils/md_lint_config.yml'
+    - name: Markdown link check
+      # cspell:words gaurav
+      uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        use-verbose-mode: 'yes'
+      if: ${{ success() || failure() }}
+    - name: "Checkout"
+      uses: actions/checkout@v2
+    - name: "Check spelling in changed files"
+      # cspell:words streetsidesoftware
+      uses: streetsidesoftware/cspell-action@v1.2.4
+      with:
+        inline: warning
+        # Only warn on spelling errors, since there's likely a lot of false
+        # positives with terminology in quantum computing.
+        strict: false
+        incremental_files_only: true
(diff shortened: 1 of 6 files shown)


.github/workflows/link_and_spell_check.yml
```diff
@@ -0,0 +1,48 @@
...
+            Write-Host "Adding matcher $_...";
+            Write-Host "::add-matcher::$_";
+          }
+      shell: pwsh
+    - name: markdownlint-cli
+      # cspell:words nosborn
+      uses: nosborn/github-action-markdown-cli@v2.0.0
+      with:
+        files: .
+        config_file: 'utils/md_lint_config.yml'
```
troelsfr, line 32: I do not see this file checked in?

.github/workflows/link_and_spell_check.yml
```diff
@@ -0,0 +1,48 @@
...
+      shell: bash
+    - run: |
+        Get-ChildItem (Join-Path "utils" "problem-matchers") `
+          | ForEach-Object {
+            Write-Host "Adding matcher $_...";
+            Write-Host "::add-matcher::$_";
+          }
+      shell: pwsh
+    - name: markdownlint-cli
+      # cspell:words nosborn
```
troelsfr, line 28: What does this comment mean?

.github/workflows/link_and_spell_check.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: "Check spelling in changed files"
+      # cspell:words streetsidesoftware
+      uses: streetsidesoftware/cspell-action@v1.2.4
+      with:
+        inline: warning
+        # Only warn on spelling errors, since there's likely a lot of false
+        # positives with terminology in quantum computing.
+        strict: false
+        incremental_files_only: true
+        config: 'utils/spelling_allowlist.json'
```
troelsfr, line 48: Likewise, I do not see this file checked in.

.github/workflows/link_and_spell_check.yml
```diff
@@ -0,0 +1,48 @@
+on:
```
troelsfr, line 1: In general, it is difficult to see how this flow is configured and it is not clear why certain requirements were chosen. For instance, what is the reason for choosing 80 characters instead of 100 or 130? Without a config checked in, it feels a bit like this limit was just chosen because it is the standard of some tool with no further thought to it.

.github/workflows/link_and_spell_check.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  check-markdown:
+    runs-on: ubuntu-latest
+    name: "Check spelling, linting and links"
+    steps:
+    - uses: actions/checkout@v2
+    - run: |
+        wget https://github.com/qir-alliance/.github/archive/refs/heads/main.zip -O clone.zip &&
+        unzip clone.zip && mv .github-main out && rm clone.zip &&
+        mv out/workflow-templates/utils/ utils && rm -r out
```
troelsfr, line 18: It would be great with some comments about what is happening here.

troelsfr (approved):
Formatting and spell checking on markdown files is a nice thing to have. We probably should also have a check for inappropriate words. This PR, is a good contribution in that direction

As a minor note, I would have preferred that
1. These checks were integrated into the structure that does the checks for C++ and Python. It would have been nice to be able to do `./manage stylecheck` locally to get the report of errors and `./manage stylecheck --fix-issues` to automatically fix them.
2. That it was more transparent which tools are performing the tasks and how they are configured.

However, that said, this is a good first step and we can consider integration at a later point.

.github/workflows/link_and_spell_check.yml
bettinaheim: This file is downloaded as part of the workflow. It is downloaded from the org page and updates to the formatting rules will hence be picked up by this workflow.

.github/workflows/link_and_spell_check.yml
bettinaheim: That's where we get this from: https://github.com/nosborn/github-action-markdown-cli

.github/workflows/link_and_spell_check.yml
bettinaheim: same as above

.github/workflows/link_and_spell_check.yml
bettinaheim: I think it is all in all helpful to stick with some rules across the board. The configuration hence lives on the org repo. Having a fixed limit on the line length makes it easier to see diffs when something is changed.

bettinaheim:
> Formatting and spell checking on markdown files is a nice thing to have. We probably should also have a check for inappropriate words. This PR, is a good contribution in that direction
> 
> As a minor note, I would have preferred that
> 
> 1. These checks were integrated into the structure that does the checks for C++ and Python. It would have been nice to be able to do `./manage stylecheck` locally to get the report of errors and `./manage stylecheck --fix-issues` to automatically fix them.
> 2. That it was more transparent which tools are performing the tasks and how they are configured.
> 
> However, that said, this is a good first step and we can consider integration at a later point.

If you have ideas for how to improve the workflow I am happy to hear them. This is a basic workflow that should facilitate common guidelines across all PRs. While using a template still required checking in a copy of the workflow, I think it is a decent step. We could even consider packing a proper GitHub action to simplify the checked in file further in the future.

.github/workflows/link_and_spell_check.yml
```diff
@@ -0,0 +1,48 @@
...
+name: "Check spelling, linting and links"
+
+jobs:
+  check-markdown:
+    runs-on: ubuntu-latest
+    name: "Check spelling, linting and links"
+    steps:
+    - uses: actions/checkout@v2
+    - run: |
+        wget https://github.com/qir-alliance/.github/archive/refs/heads/main.zip -O clone.zip &&
```
bettinaheim, line 16:
```suggestion
        # we download the config files from the org repo and pick up the rules from the latest main
        wget https://github.com/qir-alliance/.github/archive/refs/heads/main.zip -O clone.zip &&
```

.github/workflows/link_and_spell_check.yml
bettinaheim: I added a comment.

PR #51, opened 2021-12-10
Feature/update readme

troelsfr:


qir/qsharp/ConstArray/validation.json
```diff
@@ -0,0 +1,74 @@
+[
```
bettinaheim, line 1: I am not sure what does file does/ is used for?

bettinaheim (approved):

qir/qsharp/ConstArray/validation.json
troelsfr: This is output from testing whether or not the output IR is compliant with a given profile. This file should not be checked in and has been removed.

Issue #52, opened 2022-01-14
Qat doesn't inline a function call even with --always-inline and --profile base specified


peter-campora:
Hi, 

It seems like the --always-inline flag doesn't always force inlining in the generated QIR. I ran the following command via qat:
`./qat --always-inline --apply -S --profile base MagicStateDistillation-10X.ll > MagicStateDistillation-BaseProfile.10X.ll`
on the attached file and received the following output in MagicStateDistillation-BaseProfile.10X.ll (I changed the ll files to text files to attach them to the issue). The entry point function in the output file, `@Microsoft__Quantum__Samples__MeasureDistilledTAtDepth10InX__Interop`, calls out to  a helper function `@Microsoft__Quantum__Samples__Distill__body.4`. 

[MagicStateDistillation-10X.txt](https://github.com/qir-alliance/qat/files/7871641/MagicStateDistillation-10X.txt)
[MagicStateDistillation-BaseProfile.10X.txt](https://github.com/qir-alliance/qat/files/7871642/MagicStateDistillation-BaseProfile.10X.txt)

Let me know if I should try again with additional arguments to qat, or whether you need any more information.

swernli:
Thanks for reporting, @peter-campora! We will investigate and see if we can find out what inline fails on that example.
@troelsfr FYI.

troelsfr:
FYI, this issue should be solved in #53 - @swernli if you have time to review, then that would be much appreciated :)

swernli:
Thanks @troelsfr for working on that! I'll review today.

troelsfr:
#53 is merged and I am closing this ticket. @peter-campora feel free to reopen if this does not solve the issue for you.

peter-campora:
I'll take a look a look at the new version of qat and test it out! Thanks all! I found a new issue, so I'll see if any updates fix this new issue and if not, I'll file a new issue.

PR #53, opened 2022-01-17
Bugs/fix issue 52

troelsfr:
This PR fixes issue #52.


swernli (approved):
Looks good! One change this enables is that O2 and O3 optimization  now imply the inliner pass. It might be worth calling that out in the documentation.

PR #57, opened 2022-02-09
Validation Unit Tests

troelsfr:
This PR introduces unit tests for profile validation alongside better logging facilities.


qir/qat/ValidationPass/ValidationPassConfiguration.cpp
```diff
@@ -0,0 +1,135 @@
...
+    String ValidationPassConfiguration::profileName() const
+    {
+        return profile_name_;
+    }
+
+    ValidationPassConfiguration ValidationPassConfiguration::fromProfileName(String const& name)
+    {
+
+        auto profile = ValidationPassConfiguration();
+        if (name == "generic")
```
swernli, line 31: Is this generic profile only for testing purposes? Or is there a production scenario associated with it?

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
```diff
@@ -0,0 +1,135 @@
...
+            profile.allow_primitive_return_   = true;
+            profile.allow_struct_return_      = true;
+            profile.allow_pointer_return_     = true;
+        }
+        else if (name == "base")
+        {
+            profile.allow_internal_calls_     = false;
+            profile.allowlist_external_calls_ = true;
+            profile.allowlist_opcodes_        = true;
+            profile.opcodes_        = Set{"br", "call", "unreachable", "ret", "inttoptr"}; //  "phi", "select",
```
swernli, line 50: We can get into the details of refining the base profile later, I'm really focusing on the infrastructure for this review. Once we have an actual base profile definition, then we can adjust this list and consider removing the limited external calls below.

qir/qat/ValidationPass/ValidationPass.cpp
```diff
@@ -12,93 +12,280 @@ namespace microsoft
...
+            else
+            {
+                ++it->second;
+            }
+
+            pointer_location_[name].push_back(current_location_);
+        }
+    }
+
+    bool ValidationPass::enforceOpcodeRequirements()
```
swernli, line 131: I tend to prefer names for functions that return boolean to have a verb that matches the boolean intent, so that it matches well with if-statements. I think something like `passesOpcodeRequirements` would be more straightforward here. As it stands, the `enforce` name could be a bit confusing in terms of whether expected return is `true` or `false` in the success case.

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
```diff
@@ -0,0 +1,135 @@
...
+                "__quantum__qis__cnot__body",  "__quantum__qis__cz__body", "__quantum__qis__cx__body",
+                "__quantum__qis__cy__body",    "__quantum__qis__h__body",  "__quantum__qis__s__body",
+                "__quantum__qis__s__adj",      "__quantum__qis__t__body",  "__quantum__qis__t__adj",
+                "__quantum__qis__x__body",     "__quantum__qis__y__body",  "__quantum__qis__z__body",
+                "__quantum__qis__rx__body",    "__quantum__qis__ry__body", "__quantum__qis__rz__body",
+                "__quantum__qis__reset__body", "__quantum__qis__mz__body", "__quantum__qis__read_result__body",
+
+            };
+            profile.allowlist_pointer_types_ = true;
+            profile.allowed_pointer_types_   = {"i8*",    "i16*",    "i32*",    "i64*",
```
swernli, line 61: Out of curiousity, what operations end up using `i16*`? I wasn't aware we had any.

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
```diff
@@ -0,0 +1,173 @@
...
+  call void @__quantum__qis__x__body(%Qubit* %6)
+  call void @__quantum__qis__x__body(%Qubit* %5)
+  call void @__quantum__qis__x__body(%Qubit* %0)
+
+  )script");
+}
+
+TEST(CirqPositive, DISABLED_BernsteinVazirani)
+{
+    // m operations is having the wrong signature
```
swernli, line 129: This test could be enabled if new variables `%9` through `%17` were declared with values like the qubits (ie: `inttoptr i64 0 to %Result*`) and the calls to measurement updated to the pattern `call void @__quantum__qis__mz__body(%Qubit* 0, %Result* 0)`. The declaration on like 36 would need to be updated too.

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
```diff
@@ -0,0 +1,173 @@
...
+  call void @__quantum__qis__h__body(%Qubit* %5)
+  call void @__quantum__qis__cx__body(%Qubit* %4, %Qubit* %2)
+  call void @__quantum__qis__h__body(%Qubit* %0)
+  call void @__quantum__qis__h__body(%Qubit* %4)
+  call void @__quantum__qis__m__body(%Qubit* %1)
+  call void @__quantum__qis__m__body(%Qubit* %3)
+  call void @__quantum__qis__m__body(%Qubit* %6)
+  call void @__quantum__qis__m__body(%Qubit* %5)
+  call void @__quantum__qis__m__body(%Qubit* %7)
+  call void @__quantum__qis__m__body(%Qubit* %0)
```
swernli, line 168: Qubit `%0` is measured twice, the second one can be removed.

qir/qat/CMakeLists.txt
```diff
@@ -14,9 +14,11 @@ microsoft_add_library_tests(Profile)
 microsoft_add_library(Validator)
 target_link_libraries(Validator PRIVATE Logging)
 
-microsoft_add_library_tests(Validator ValidationPass Logging)
+microsoft_add_library_tests(Validator  Rules TransformationRulesPass AllocationManager Commandline Generators Validator ValidationPass Logging)
 microsoft_add_library(ValidationPass Logging)
 
+
```
swernli, line 20: These extra spaces can be removed.

qir/qsharp/NegativeBaseProfiles/ExternalCall/QSharpVersion/Example.qs
```diff
@@ -0,0 +1,16 @@
+namespace LoopCase {
```
swernli, line 1: Calling this `LoopCase` is a bit confusing since it doesn't involve any loops.

qir/qsharp/NegativeBaseProfiles/ExternalCall/QSharpVersion/Example.qs
```diff
@@ -0,0 +1,16 @@
...
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Measurement;
+
+    @EntryPoint()
+    operation Main(nrIter: Int): Int
```
swernli, line 6: This function doesn't seem to correlate to the QIR that is tested in the QSharpTypeQirsNegative.cpp tests. Is this leftover from an unused test case?

qir/qsharp/NegativeBaseProfiles/InternalCall/QSharpVersion/Example.qs
```diff
@@ -0,0 +1,16 @@
+namespace LoopCase {
```
swernli, line 1: Here too, `LoopCase` seems to be the wrong name.

qir/qsharp/NegativeBaseProfiles/IfStatement/QSharpVersion/Example.qs
```diff
@@ -0,0 +1,18 @@
+namespace IfCase {
```
swernli, line 1: Here, the Q# uses the accurate name `IfCase` but the test that seems to correlate uses the name `IfWithPhi` and QIR snippet that has `LoopCase` in the function name. This could cause someone trouble later if they try to use this file to regenerate the QIR for that test case.

swernli (commented):
I've left a few questions and feedback. Overall the infrastructure looks great!

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
troelsfr: I am not sure we use it anywhere, but neither do I think we use `i32*` and `i64*`. However, it seems reasonable to allow pointers to all of these primitive types if you allow pointers to one given that we should not base the profile on what Q# currently does, but rather think of it as something that will be used by many different frontends.

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
troelsfr: This test is actually disabled for the time being. I have to fix my ad-hoc cirq converter or wait for the real one to arrive to get the code right here.

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
troelsfr: Where do you see the first measurement?

qir/qsharp/NegativeBaseProfiles/ExternalCall/QSharpVersion/Example.qs
troelsfr: Except that `nrIter` and `ret` is not used, I see a pretty good correlation with the function in the test:

```llvm
  %q.i = tail call %Qubit* @__quantum__rt__qubit_allocate()
  tail call void @__quantum__qis__h__body(%Qubit* %q.i)
  %result.i.i = tail call %Result* @__quantum__qis__m__body(%Qubit* %q.i)
  tail call void @__quantum__qis__reset__body(%Qubit* %q.i)
```
and
```q#
        use q = Qubit();
        H(q);
        let _ = MResetZ(q);
```
Am I missing something here?

qir/qsharp/NegativeBaseProfiles/ExternalCall/QSharpVersion/Example.qs
troelsfr: Updated to match the name in the test.

qir/qsharp/NegativeBaseProfiles/InternalCall/QSharpVersion/Example.qs
troelsfr: Name updated.

qir/qsharp/NegativeBaseProfiles/IfStatement/QSharpVersion/Example.qs
troelsfr: Name updated.

qir/qat/TestTools/IrManipulationTestHelper.hpp
```diff
@@ -55,10 +55,32 @@ namespace quantum
...
             bool                     debug              = false);
 
+        /// Validates a profile to the module to allow which transforms the IR. This
+        /// allow us to write small profiles to test a single piece of transformation.
+        bool validateProfile(GeneratorPtr const& generator, String const& profile_name = "generic", bool debug = false);
+
+        /// Tests whether a given set of errors (LLVM hints) are present in the validation errors
+        /// for a specific profile. This method only checks if errors are present but does not fail if
+        /// there are more errors than requested through the API.
+        bool hasValidationErrors(
```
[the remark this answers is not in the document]
troelsfr: Renamed.

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
troelsfr: At the moment we do not have a production scenario other than you might want to bypass validation and this can be done by choosing a generic profile that allows everything. Later on we may want to add such things a name enforcing which is also done for the generic case.

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
troelsfr: Yes, I agree. This is not meant to be the final version, but it will be revised as we go.

qir/qat/ValidationPass/ValidationPass.cpp
troelsfr: I've changed it to `satisfiesOpcodeRequirements`

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
swernli: Oh, I misunderstood this profile and thought it was a deny all. Allow all makes sense as something we'd need for testing and possible future user scenarios. Thanks for explaining!

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
troelsfr: I've removed this code.

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
swernli: It's measured first on line 161 then again here on line 168. Could be a bug in the high level code that was used to generate this QIR, but given you already said this test case is going to stay disabled then it's not a big deal.

qir/qsharp/NegativeBaseProfiles/ExternalCall/QSharpVersion/Example.qs
swernli: Ah, now that you've renamed it I can see which test case it corresponds to, I didn't notice it before. Thanks!

swernli (approved):

qir/qat/Validator/Tests/Unit/CirqTypeQirsPositive.cpp
troelsfr: Line 161 is an H gate, not measure :) I will leave it as it is as it is as this part is likely to be updated when the converter is done.

PR #58, opened 2022-02-23
Moving features into experimental

troelsfr:
All experimental and non-experimental features are listed in the help section:
```
% ./qir/qat/Apps/qat
Usage: /Users/tfr/Documents/Projects/qat/Debug/qir/qat/Apps/qat [options] filename

Component configuration - Used to disable or enable components

--disable-validation-configuration                Disables Validation configuration. Default: false
--disable-llvm-optimisation                       Disables LLVM optimisations. Default: false
--disable-transformation-rules                    Disables Pass configuration. Default: false
--disable-grouping                                Disables Grouping quantum instructions. Default: false

Base configuration - Configuration of the quantum adoption tool to execute a specific behaviour.

--load                                            EXPERIMENTAL. Load component. Default:
--apply                                           Applies a profile to transform the IR in correspondence with the profile. Default: false
--validate                                        Executes the validation procedure. Default: false
--profile                                         Sets the profile. Default: generic
--S                                               Emits LLVM IR to the standard output. Default: false
--O0                                              Optimisation level 0. Default: false
--O1                                              Optimisation level 1. Default: false
--O2                                              Optimisation level 2. Default: false
--O3                                              Optimisation level 3. Default: false
--verify-module                                   Verifies the module after transformation. Default: false
--experimental                                    Enables experimental features. Default: false
--dump-config                                     Prints the configuration to the standard output. Default: false

Transformation rules - Rules used to transform instruction sequences in the QIR.

--[no-]optimise-result-one                        EXPERIMENTAL. Maps branching based on quantum measurements compared to one to base profile type measurement. Default: true
--[no-]optimise-result-zero                       EXPERIMENTAL. Maps branching based on quantum measurements compared to zero to base profile type measurement. Default: true
--[no-]use-static-qubit-array-allocation          Maps allocation of qubit arrays to static array allocation. Default: true
--[no-]use-static-qubit-allocation                Maps qubit allocation to static allocation. Default: true
--[no-]use-static-result-allocation               Maps result allocation to static allocation. Default: true
--[no-]disable-reference-counting                 Disables reference counting by instruction removal. Default: true
--[no-]disable-alias-counting                     Disables alias counting by instruction removal. Default: true
--[no-]disable-string-support                     Disables string support by instruction removal. Default: true

Validation configuration -

--[no-]allow-internal-calls                       Whether or not internal calls are allowed. Default: true
--save-validation-report                          Saves the validation report to specified filename. Default:

LLVM optimisations - Enables specific LLVM optimisations before and after transformation.

--unroll-loops                                    Aggressively unroll loops. Default: false
--[no-]allow-partial                              Enables or disables partial unrolling. When disabled only full unrolling is allowed. Default: true
--[no-]allow-peeling                              Enables or disables loop peeling. Default: true
--[no-]allow-runtime                              Enables or disables unrolling of loops with runtime trip count. Default: true
--[no-]allow-upper-bound                          Enables or disables the use of trip count upper bound in loop unrolling. Default: true
--[no-]allow-profile-based-peeling                Enables or disables loop peeling basing on profile. Default: true
--full-unroll-count                               Sets the max full unroll count. Default: 1024
--unroll-opt-level                                Sets the optimisation level for loop unrolling. Default: 3
--only-when-forced                                If true, only loops that explicitly request unrolling via metadata are considered. Default: false
--forget-scev                                     If true, forget all loops when unrolling. If false, forget top-most loop of the currently processed loops. Default: false
--always-inline                                   Aggressively inline function calls. Default: false
--inlining-parameter                              Number of code lines acceptable when inlining. Default: 2147483647
--[no-]eliminate-constants                        Uses LLVM pass to eliminate constants. Default: true
--[no-]eliminate-dead-code                        Uses LLVMs aggressive dead code elimination. Default: true
--[no-]eliminate-memory                           Maps memory into registers where feasible. Default: true
--use-llvm-opt-pipeline                           EXPERIMENTAL. Disables the the default pipeline. Default: false
--opt-pipeline-config                             EXPERIMENTAL. LLVM passes pipeline to use upon applying this component. Default:

Pass configuration - Configuration of the pass and its corresponding optimisations.

--[no-]delete-dead-code                           EXPERIMENTAL. Deleted dead code. Default: true
--[no-]clone-functions                            EXPERIMENTAL. Clone functions to ensure correct qubit allocation. Default: true
--[no-]transform-execution-path-only              EXPERIMENTAL. Transform execution paths only. Default: true
--max-recursion                                   EXPERIMENTAL. Defines the maximum recursion when unrolling the execution path Default: 512
--assume-no-except                                EXPERIMENTAL. Assumes that no exception will occur during runtime. Default: false
--reuse-qubits                                    EXPERIMENTAL. Use to define whether or not to reuse qubits. Default: false
--reuse-results                                   EXPERIMENTAL. Use to define whether or not to reuse results. Default: false
--entry-point-attr                                Specifies the attribute indicating the entry point. Default: InteropFriendly
--[no-]annotate-qubit-use                         Annotate the number of qubits used Default: true
--[no-]annotate-result-use                        Annotate the number of results used Default: true

Grouping quantum instructions -

--[no-]separate-circuits                          EXPERIMENTAL. Whether or not to separate quantum and classical circuits Default: true
```


bettinaheim:
@troelsfr Could you please ensure that all APIs (and also doc comments, but I don't mind if these get gradually updated) follow American standard spelling? While that is not a reflection of who the tool is for, overall it is nicer and less surprising for users to have some consistency across all out tools.

troelsfr:
Sure. I will pick this up on Monday.

swernli (commented):
Code changes look good, but one thing that took me a while to piece together is that experimental features effectively have two default settings: one default when experimental is not provided and one when it is. The help description of these parameters is a little confusing in that regard, because it describes some parameters as default `true` and it makes it seem as if some experimental features are enabled by default for normal operation. Perhaps some additional text on the `--experimental` parameter itself to clarify that any features that are marked `EXPERIMENTAL` are disabled if the top level `--experimental` param is not provided?

troelsfr:
> Code changes look good, but one thing that took me a while to piece together is that experimental features effectively have two default settings: one default when experimental is not provided and one when it is. The help description of these parameters is a little confusing in that regard, because it describes some parameters as default `true` and it makes it seem as if some experimental features are enabled by default for normal operation. Perhaps some additional text on the `--experimental` parameter itself to clarify that any features that are marked `EXPERIMENTAL` are disabled if the top level `--experimental` param is not provided?

Yes, this is a good suggestion. I will make this change today.

swernli (approved):

PR #66, opened 2022-03-23
Adding support for IR debug info

troelsfr:
Using
```c++
#include <iostream>
#include <vector>

int main()
{
  std::vector<int> x;
  std::cout << "Hello world" << std::endl;
  x[10299] = x[10299 * 2777] + 2;
  std::cout << "X[10299] = " << x[10299] << std::endl;
  return 0;
}
```
We create an IR and use QAT to add debug info
```
% clang++ -O3 -c -S -emit-llvm segfault.cpp
% qat -S segfault.ll > with_debug.ll
% mkdir -p bin
% llc -filetype=obj -o bin/testprog.o with_debug.ll
% clang++ -g -O1 bin/testprog.o -o bin/testprog
% lldb bin/testprog
```
Executing the program in the debugger we now see that the IR file is referred to when the segfault is happening:
```
(lldb) target create "bin/testprog"
Current executable set to '/Users/tfr/Documents/TestArea/LLVMIR/bin/testprog' (x86_64).
(lldb) r
Process 88015 launched: '/Users/tfr/Documents/TestArea/LLVMIR/bin/testprog' (x86_64)
Hello world
Process 88015 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x6d1a00c)
    frame #0: 0x0000000100003a4e testprog`main at segfault.ll:72:75
   69  	  call void @llvm.lifetime.end.p0i8(i64 8, i8* nonnull %11) #9
   70  	  %24 = call nonnull align 8 dereferenceable(8) %"class.std::__1::basic_ostream"* @_ZNSt3__113basic_ostreamIcNS_11char_traitsIcEEE3putEc(%"class.std::__1::basic_ostream"* nonnull %3, i8 signext %20)
   71  	  %25 = call nonnull align 8 dereferenceable(8) %"class.std::__1::basic_ostream"* @_ZNSt3__113basic_ostreamIcNS_11char_traitsIcEEE5flushEv(%"class.std::__1::basic_ostream"* nonnull %3)
-> 72  	  %26 = load i32, i32* inttoptr (i64 114401292 to i32*), align 4, !tbaa !6
   73  	  %27 = add nsw i32 %26, 2
   74  	  store i32 %27, i32* inttoptr (i64 41196 to i32*), align 4, !tbaa !6
   75  	  %28 = call nonnull align 8 dereferenceable(8) %"class.std::__1::basic_ostream"* @_ZNSt3__124__put_character_sequenceIcNS_11char_traitsIcEEEERNS_13basic_ostreamIT_T0_EES7_PKS4_m(%"class.std::__1::basic_ostream"* nonnull align 8 dereferenceable(8) @_ZNSt3__14coutE, i8* getelementptr inbounds ([12 x i8], [12 x i8]* @.str.1, i64 0, i64 0), i64 11)
(lldb)
```

### DIFF
--- a/qir/qat/ModuleLoader/DebugTable.cpp
+++ b/qir/qat/ModuleLoader/DebugTable.cpp
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "ModuleLoader/DebugTable.hpp"
+
+#include "Llvm/Llvm.hpp"
+
+#include <memory>
+
+namespace microsoft {
+namespace quantum {
+
+DebugTable::DebugTablePtr DebugTable::create()
+{
+  DebugTablePtr ret;
+  ret.reset(new DebugTable());
+  return ret;
+}
+
+void DebugTable::printInfoComment(Value const &value, llvm::formatted_raw_ostream &outstream)
+{
+  registerValuePosition(&value, outstream);
+}
+
+void DebugTable::emitBasicBlockStartAnnot(BasicBlock const            *block,
+                                          llvm::formatted_raw_ostream &outstream)
+{
+  registerValuePosition(block, outstream);
+}
+
+void DebugTable::emitFunctionAnnot(Function const *function, llvm::formatted_raw_ostream &outstream)
+{
+  registerValuePosition(function, outstream);
+}
+
+DebugTable::Position DebugTable::getPosition(Value const *value) const
+{
+  auto it = positions_.find(value);
+  if (it != positions_.end())
+  {
+    return it->second;
+  }
+
+  return Position::InvalidPosition();
+}
+
+void DebugTable::registerModule(StringRef const &filename, Module const *module)
+{
+  current_filename_ = filename;
+  llvm::raw_null_ostream dummy{};
+  module->print(dummy, this);
+}
+
+}  // namespace quantum
+}  // namespace microsoft
--- a/qir/qat/ModuleLoader/ModuleLoader.hpp
+++ b/qir/qat/ModuleLoader/ModuleLoader.hpp
@@ -2,132 +2,167 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#include "Llvm/Llvm.hpp"
+#include "ModuleLoader/DebugInfoUpdater.hpp"
+#include "ModuleLoader/DebugTable.hpp"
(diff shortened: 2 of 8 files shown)


qir/qat/ModuleLoader/DebugTable.cpp
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        return Position::InvalidPosition();
+    }
+
+    void DebugTable::registerModule(StringRef const& filename, Module const* module)
+    {
+        current_filename_ = filename;
+        llvm::raw_null_ostream dummy{};
+        module->print(dummy, this);
```
swernli, line 51: Since printing to the null stream using the table as the annotation writer has the effect of capturing info about the module, it might  be worth a clarifying comment here. At first glance the use of the null stream can be confusing.

qir/qat/ModuleLoader/ModuleLoader.hpp
```diff
@@ -38,33 +42,74 @@ namespace quantum
...
+                auto debug_version_key = "Debug Info Version";
+                if (!module->getModuleFlag(debug_version_key))
+                {
+                    module->addModuleFlag(llvm::Module::Warning, debug_version_key, llvm::DEBUG_METADATA_VERSION);
+                }
+
+                // Update with debug information
+                DebugInfoUpdater updater(debug_table_, *module.get(), directory, filename);
+                updater.update();
+            }
```
swernli, line 93: I'm a little confused by the intent of `strip_existing_debug_` here. The existing debug information is unconditionally captured into the table, then the info is optionally stripped from the module. This means the strip option doesn't suppress the existing debug info from being present during later feedback. So is the debug table and it's use meant to be independent of the strip option or could there be an ordering issue here?

qir/qat/ModuleLoader/DebugTable.cpp
troelsfr: That is a good point.

qir/qat/ModuleLoader/ModuleLoader.hpp
troelsfr: `strip_existing_debug_` refers to the debug symbols that existed _before_ the IR was loaded by QAT. Notice that interchanging the strip-if block and line 71 does not change the meaning of the program. I think the confusion is that `debug_table_` really is a `location_table_`. I will change the naming.

swernli (approved):

PR #70, opened 2022-03-24
Q# Tests

troelsfr:
This PR introduces tests for Q# samples in https://github.com/microsoft/Quantum/. This is used as a means to ensure that all samples are correctly loaded and reduce (when possible) without severe failures.


qir/qsharp-tests/qsharp_tests/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+                qir_generation = ET.SubElement(group, 'QirGeneration')
+            qir_generation.text = "true"
+
+            library_type = group.find("OutputType")
+            if library_type is not None:
+                library_type.text = "Exe"
+
+        for group in root.iter('ItemGroup'):
+            if group.find("ProjectReference") is not None:
+                self.has_relative_dependencies = True
```
swernli, line 48: In theory, QIR generation for projects with relative dependencies should work, but I agree that scoping it this way for now is a good way to focus on a subset of samples for the time being (and a neat mechanism too).

qir/qsharp-tests/test_qsharp_validation.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.mark.parametrize("test_name", CIRCUITS)
+def test_qsharp_reduction(test_name, request):
+    with request.getfixturevalue(test_name) as project:
+        if project.has_relative_dependencies:
+            return
+
+        if not project.compile():
+            logger.warn("Q# program ignored as it could not compile as a library.")
```
swernli, line 67: nitpick: this looks like the string might be leftover from the an earlier approach, since everything is compiled as an exe now.

swernli (approved):
Approved with some minor feedback. This infrastructure is going to be very useful going forward!

qir/qsharp-tests/qsharp_tests/__init__.py
troelsfr: The reason why I have to skip the ones with relative dependence is that I copy each project out to a temporary folder before modifying it such that you can use it with your local setup without destroying the repo. In principle, you could fix the relative dependence by referring back to the original repo, but it seems like a lot of work and very little benefit.  I am happy to review this at a later point though.

qir/qsharp-tests/test_qsharp_validation.py
troelsfr: No, this is supposed to be there such that all tests that actually compile run but you get informed about those projects were not included. I.e. we don't want the exit to be an error.

PR #76, opened 2022-04-18
Updating Qubit annotation

troelsfr:


qir/qat/StaticResourcePass/StaticResourcePass.cpp
```diff
@@ -0,0 +1,204 @@
...
+            }
+        }
+
+        ret.usage_qubit_counts  = qubits_used.size();
+        ret.usage_result_counts = results_used.size();
+
+        return ret;
+    }
+
+    void StaticResourcePass::annotateQubits(llvm::Module& module) const
```
swernli, line 140: I think the intention we want to go for is that a transformation pass that changes qubit indexing (either by switching to global indexing from dynamic allocation or as part of a pass that might remap the identifiers) should update the `requiredQubits` annotation. In this case, because this pass does not remap the indices to be sequential without gaps, the required qubits counting can get out of sync with the identifiers used. The pass should remap the qubit identifiers (which may or may not actually result in changes, depending on the initial state) and then update the `requriedQubits` attribute, or be disabled and then leave any existing attributes alone. The combination of remapping plus counting uses into `requiredQubits` should replace the need for a `maxQubitIndex` attribute.

qir/qat/StaticResourcePass/StaticResourcePassConfiguration.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+namespace quantum
+{
+
+    void StaticResourcePassConfiguration::setup(ConfigurationManager& config)
+    {
+        config.setSectionName("Static resource manipulation", "");
+        config.addParameter(annotate_qubit_use_, "annotate-qubit-use", "Annotate the number of qubits used");
+
+        config.addParameter(annotate_result_use_, "annotate-result-use", "Annotate the number of results used");
```
swernli, line 18: Rather than having one pass that transforms dynamic allocation into global identifiers and does not add the attribute followed by a pass that performs counting of resources and adds the attributes, I think we should have the following design:

- A pass that transforms dynamic allocation to global identifiers AND adds the `requiredQubits` attribute with a value consistent with the indices of qubits used.
- A pass that may remap global qubit identifiers into a contiguous order starting with `0` and updates/adds the `requiredQubits` attribute with a value consistent with the number of qubits used.

This way, a pass that may add or change any global ids is always responsible for adding/updating the attribute to match the transformation performed. If neither pass is run, any existing attribute is left alone. So presence or absence of the attribute is no longer independently controlled, but rather a program that uses static global identifiers is expected to have the attribute to act as a sort of declaration of those global resources that must  be consistent with the identifiers used to index into that global resource.

@bettinaheim does this match your expectation?

troelsfr: Calculation of the used number of qubits needs to be independent of the allocation unit. The reason for this is that optimisation passes applied _after_ removing the virtualisation (or directly on a QIR with global ids) may choose to remove redundant qubits. For instance,
```qsharp
operation T() : Bool
{
  use dummy = Qubit();
  use q1 = Qubit();
  H(q1);
  return MReset(q1) == One;
}
```
In the above program, dummy would be translated into qubit id 0 and q1 as qubit id 1. If the QIR is annotated at allocation time, the required number of qubits will be incorrect if a later pass would optimise the redundant variable `dummy` out of the system. Hence, I would advise against having `requiredQubits` managed by the allocating entity and rather something that is done as an after-thought with optional remapping.

qir/qat/StaticResourcePass/StaticResourcePass.cpp
troelsfr: I've implemented a first version of remapping. I believe that maxQubitIndex still may be a useful feature, so I would suggest to just make this annotation off by default.

qir/qat/StaticResourcePass/StaticResourcePassConfiguration.cpp
swernli: I think we are still not quite on the same page here. What I'm advocating for is that qubit counting is not a separate pass, but rather a common feature that any pass can (and should) call into after a transformation that can add or change qubit global ids. So the dyanmic allocation to global id transformation would call into counting after transforming and get a count in the attribute that is accurate a that time. Then remapping would also call into the counting code to get an updated count after having changed the global identifiers. What I'd like to assert is this: if a program has global identifiers then it has the required qubit attribute AND that count should always be up-to-date relative to the last transformation.

swernli: This is something that we can refine going forward. For now, this unblocks us as long as we use the initial remapping pass introduced here.

swernli (approved):
Approving so we can unblock further testing.

qir/qat/StaticResourcePass/StaticResourcePassConfiguration.cpp
troelsfr: I believe I understand what you are saying and I have a couple of points to make in that regard:

1. The way the LLVM infrastructure is built, you do not have a "component" that passes can call into. If a pass relied on needing to know the number of qubits, such information would be separated into an analysis pass that could be reused by other passes. To update an attribute, you would either update the attribute in each individual pass or make a new pass that updates the attribute.

2. For that reason, it is highly unlikely that we would ever have a pass that would depend on this attribute. As I see it, this attribute is solely for the non-qat world.

3. We can remove the parameter to specify whether the counting qubit part of the pass should be invoked. However, this would prevent pass-through of this attribute where an external program would be dictating the value (i.e. PyQir for instance). I am fine either way and it is not clear to me whether there are any use cases where you would want to do this although I understood that there are cases where it is preferred PyQir to be in charge.

4. We could add auto-detection of static/dynamic allocation. and have the flag set on the basis of this. This would require an additional analysis pass. However, it does leave the question what do you do in the case of mixed allocation? Is this allowed? Is it banned? And is it a QAT responsibility to enforce a ban?

PR #77, opened 2022-04-18
Update additional profile, add record output pruning

swernli:
This adds details for the thrid profile configured for validation.

It also adds a new option disabling record output support, which will prune all of the `__quantum__rt__*_record_output` function calls. This is on by default to match the behavior of other runtime support functions, and can be turned off via `--no-disable-record-output-support`.

### DIFF
--- a/.github/workflows/trusted-ci-mac-on-main-pr.yml
+++ b/.github/workflows/trusted-ci-mac-on-main-pr.yml
@@ -83,6 +83,9 @@ jobs:
         run: |
           git clone git@github.com:msr-quarc/qiskit-qir.git
           pushd qiskit-qir
+
+          git checkout 111e8149e5ea79c20764c044c454facc5188b2bd
+
           pip3 install --user -e .
           popd
           export PYTHONPATH=$PWD/qiskit-qir/tests:$PYTHONPATH
--- a/.github/workflows/trusted-ci-mac-on-main-pr.yml
+++ b/.github/workflows/trusted-ci-mac-on-main-pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: "Copying QAT source"
         run: |
           cp -R qir/qat /tmp/
-          cp -R qir/qiski-tests /tmp/
+          cp -R qir/qiskit-tests /tmp/
 
       - name: Checkout code main
         uses: actions/checkout@v2
--- a/.github/workflows/trusted-ci-mac-on-main-pr.yml
+++ b/.github/workflows/trusted-ci-mac-on-main-pr.yml
@@ -25,6 +25,7 @@ jobs:
       - name: "Copying QAT source"
         run: |
           cp -R qir/qat /tmp/
+          cp -R qir/qiski-tests /tmp/
 
       - name: Checkout code main
         uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


qir/qat/ValidationPass/ValidationPassConfiguration.cpp
```diff
@@ -67,9 +67,9 @@ namespace quantum
             profile.allow_internal_calls_     = false;
             profile.allowlist_external_calls_ = true;
             profile.allowlist_opcodes_        = true;
-            profile.opcodes_                  = Set{"ret",  "inttoptr", "br",   "phi", "add", "sub", "mul",  "fadd",
-                                   "fsub", "fmul",     "ashr", "and", "or",  "xor", "icmp", "fcmp"};
-            profile.external_calls_           = Set{
+            profile.opcodes_        = Set{"call", "ret",  "inttoptr", "br",  "phi", "add", "sub",  "mul", "fadd",
```
cesarzc, line 70: Are these changes to the allowed opcodes for the other profiles also meant to be part of this PR?

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
```diff
@@ -67,9 +67,9 @@ namespace quantum
...
+                                   "fsub", "fmul", "ashr",     "and", "or",  "xor", "icmp", "fcmp"};
```
cesarzc, line 71: nit: spacing before the `"and"`

swernli: Turns out that clang-format enforces this... without the spacing it will fail CI

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
swernli: I did include this intentionally, yes. It was an oversight to leave "call" out of the list before.

qir/qat/ValidationPass/ValidationPassConfiguration.cpp
cesarzc: That seems like an unexpected clang-format enforcement :P

cesarzc (approved):

.github/workflows/trusted-ci-mac-on-main-pr.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       - name: "Copying QAT source"
         run: |
           cp -R qir/qat /tmp/
+          cp -R qir/qiskit-tests /tmp/
```
troelsfr, line 28:
```suggestion
```

.github/workflows/trusted-ci-mac-on-main-pr.yml
```diff
@@ -35,7 +36,9 @@ jobs:
       - name: "Replacing QAT source"
         run: |
           rm -rf qir/qat
+          rm -rf qir/qiskit-tests
```
troelsfr, line 39:
```suggestion
```

.github/workflows/trusted-ci-mac-on-main-pr.yml
```diff
@@ -35,7 +36,9 @@ jobs:
...
           mv /tmp/qat qir/
+          mv /tmp/qiskit-tests qir/
```
troelsfr, line 41:
```suggestion
```

troelsfr (approved):
I would suggest that we remove the copy of `qiskit-tests`.

swernli:
@troelsfr I've added the pruning of the new runtime functions as default behavior, so I re-requested your review. Let me know if anything in there looks amiss, and either you or I can merge the change if you sign off again!

troelsfr:
@swernli it looks good to me.

PR #81, opened 2022-05-02
Improved diagnostics and code refactor

troelsfr:
This PR introduces improved diagnostics and pre validation of input code to ensure that it is compliant with requirements for the selected flags. To accomplish this, this PR introduces deferred configuration flags which can be referenced by other configurations to determine how they operate. The pre-validation pass ensures that only straight-line code is accepted in the case where we request QAT to replace qubits after resets. This PR also substantially improves the error reporting and removes many `stderr` messages.

Amongst minor updates, it deprecates the name "base" profile and introduces "default" profile instead.

### DIFF
--- a/qir/qat/Apps/Qat/Qat.cpp
+++ b/qir/qat/Apps/Qat/Qat.cpp
@@ -322,6 +322,12 @@ int main(int argc, char** argv)
             logger->dump(fout);
             fout.close();
         }
+
+        // Enforcing that logged errors causes non-zero return value.
+        if (logger && (logger->hadErrors() || logger->hadWarnings()))
+        {
+            ret = -1;
+        }
     }
     catch (std::exception const& e)
     {
--- a/qir/qat/Apps/Qat/Qat.cpp
+++ b/qir/qat/Apps/Qat/Qat.cpp
@@ -281,20 +281,29 @@ int main(int argc, char** argv)
 
         auto profile = generator->newProfile(config.profile(), optimization_level, config.isDebugMode());
 
-        if (config.shouldGenerate())
+        if (ret == 0 && config.shouldGenerate())
         {
             profile.apply(*module);
+
+            //  Preventing subsequent routines to run if errors occurred.
+            if (logger && (logger->hadErrors() || logger->hadWarnings()))
+            {
+                ret = -1;
+            }
         }
 
         // We deliberately emit LLVM prior to verification and validation
         // to allow output the IR for debugging purposes.
-        if (config.shouldEmitLlvm())
+        if (ret == 0)
         {
-            llvm::outs() << *module << "\n";
-        }
-        else
-        {
-            llvm::WriteBitcodeToFile(*module, llvm::outs());
+            if (config.shouldEmitLlvm())
+            {
+                llvm::outs() << *module << "\n";
+            }
+            else
+            {
+                llvm::WriteBitcodeToFile(*module, llvm::outs());
+            }
         }
 
         if (ret == 0 && config.verifyModule())
@@ -302,6 +311,13 @@ int main(int argc, char** argv)
             if (!profile.verify(*module))
             {
                 std::cerr << "IR is broken." << std::endl;
+                ret = -1;
+            }
+
+            // Safety pre-caution to ensure that all errors and warnings reported
+            // results in failure.
+            if (logger && (logger->hadErrors() || logger->hadWarnings()))
+            {
                 ret = -1;
             }
         }
@@ -321,12 +337,6 @@ int main(int argc, char** argv)
             std::fstream fout(config.saveReportTo(), std::ios::out);
             logger->dump(fout);
             fout.close();
-        }
-
-        // Enforcing that logged errors causes non-zero return value.
-        if (logger && (logger->hadErrors() || logger->hadWarnings()))
-        {
-            ret = -1;
         }
     }
     catch (std::exception const& e)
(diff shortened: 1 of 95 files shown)


qir/qat/Apps/Qat/Qat.cpp
```diff
@@ -281,20 +281,29 @@ int main(int argc, char** argv)
 
         auto profile = generator->newProfile(config.profile(), optimization_level, config.isDebugMode());
 
-        if (config.shouldGenerate())
+        if (ret == 0 && config.shouldGenerate())
```
kuzminrobin, line 284: I have an impression that `ret` is always `0` here.

qir/qat/Apps/Qat/QatConfig.cpp
```diff
@@ -48,6 +48,12 @@ namespace quantum
 
     String QatConfig::profile() const
     {
+        if (profile_ == "base")
+        {
+            // TODO(tfr): Remove warning upon final release.
+            llvm::errs() << "; WARNING: 'base' profile renamed 'default'. Please update your scripts.\n";
+            return "default";
+        }
```
kuzminrobin, line 56:
```suggestion
        if (profile_ == "base")
        {
            // TODO(tfr): Remove warning upon final release.
            llvm::errs() << "; WARNING: 'base' profile renamed to 'default'. Please update your scripts.\n";
            return "default";
        }
```

qir/qat/Commandline/ConfigurationManager.hpp
```diff
@@ -21,6 +21,53 @@ namespace microsoft
...
+
+        void setReference(IConfigBindPtr const& value)
+        {
+            value_ref_ = value;
+        }
+
+        bool isDereferencable() const
+        {
+            return value_ref_ != nullptr;
+        }
```
kuzminrobin, line 47: Nit.
```suggestion
        bool isDereferenceable() const
        {
            return value_ref_ != nullptr;
        }
```
[Details](https://en.wiktionary.org/wiki/referenceable).

qir/qat/Commandline/ConfigurationManager.hpp
```diff
@@ -21,6 +21,53 @@ namespace microsoft
...
+
+        template <typename T> T value() const
+        {
+            if (!isDereferencable())
+            {
+                throw std::runtime_error("Deferred command line parameter is not dereferrencable.");
```
kuzminrobin, line 53:
```suggestion
                throw std::runtime_error("Deferred command line parameter is not dereferenceable.");
```

qir/qat/GroupingPass/GroupingPass.cpp
```diff
@@ -309,7 +317,15 @@ namespace quantum
...
+                    logger_->setLocationFromValue(ptr);
+                    logger_->error("Could not delete node.");
+                }
+                else
+                {
+                    throw std::runtime_error("No logger present - Error: Unable to delete instruction.\n");
+                }
             }
             else
             {
```
kuzminrobin, line 331: Feels like 2 identical loops in green lines 249 - 268 and 315 - 334. Consider de-duplicating.

qir/qat/PreTransformValidation/PreTransformValidationPass.cpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+    }
+
+    llvm::PreservedAnalyses PreTransformValidationPass::run(llvm::Module& module, llvm::ModuleAnalysisManager& /*mam*/)
+    {
+        uint64_t function_count{0};
+
+        if (config_.requireStraightLineCode())
+        {
+            for (auto& function : module)
```
kuzminrobin, line 30: Consider moving `function_count` to inside of `if`. Otherwise, if `if` is not entered then the allocation and initialization in line 26 is redundant.
```suggestion
        if (config_.requireStraightLineCode())
        {
            uint64_t function_count{0};

            for (auto& function : module)
```

qir/qat/PreTransformValidation/PreTransformValidationPass.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+                ++function_count;
+
+                if (function_count > 1)
+                {
+                    if (logger_)
+                    {
+                        logger_->setLocationFromValue(&function.getEntryBlock());
+                        logger_->error("Expected straight line code, but multiple functions present.");
+                    }
```
kuzminrobin, line 46: I have an impression that line 45 can be executed multiple times in the loop (starting in line 30). Did you mean placing `break;` or `return;` between lines 45 and 46?

Same for line 62 below.

kuzminrobin (approved):

qir/qat/Apps/Qat/Qat.cpp
troelsfr: That is true - corrected.

qir/qat/GroupingPass/GroupingPass.cpp
troelsfr: Yes, this is a good point. This has been fixed.

qir/qat/PreTransformValidation/PreTransformValidationPass.cpp
troelsfr: The reason for this is that if a logger is present, we highlight all errors. In this case, that means all functions beyond the first one. However, in the absence of a logger, we throw to 1) create a signal that the program has failed and 2) to avoid flooding the `stderr`

swernli (approved):

PR #86, opened 2022-05-27
Moving pre validation to post validation

troelsfr:


qir/qsharp/TargetPackages/QSharpVersion/qir/Example.ll
```diff
@@ -0,0 +1,140 @@
+
```
bettinaheim, line 1: Is this file checked in intentionally? I think it doesn't need to be, given that a build would just produce it.

bettinaheim (approved):

PR #89, opened 2022-05-31
Upgrading to LLVM 13

troelsfr:


.github/workflows/ci-linux-on-main-pr.yml
```diff
@@ -26,14 +26,13 @@ jobs:
...
           sudo apt-get install -y curl pkg-config findutils  wget
-          sudo apt install -y clang-11 cmake clang-format-11 clang-tidy-11 
-          sudo apt-get install -y llvm-11 lldb-11 llvm-11-dev libllvm11 llvm-11-runtime opt 
+          sudo apt install -y clang-13 cmake clang-format-13 clang-tidy-13 
+          sudo apt-get install -y llvm-13 lldb-13 llvm-13-dev libllvm11 llvm-13-runtime opt 
           sudo apt-get install -y nodejs npm
           sudo apt install -y python3 python3-pip 
-          sudo apt-get remove -y llvm-10
-          sudo ln -s /usr/lib/llvm-11/bin/opt /usr/bin/opt
+          sudo ln -s /usr/lib/llvm-13/bin/opt /usr/bin/opt
```
kuzminrobin, line 35: Up to you.
It probably makes sense to create a constant (variable) that contains the version of active LLVM (in this case `13`), and use that constant everywhere. Otherwise there is a chance that some time later someone will be updating version 13 to 14+, will update the green lines 31 - 32, but will forget to update `13` in the green line 35.

.github/workflows/trusted-cd-linux-release.yml
```diff
@@ -20,8 +20,8 @@ jobs:
           sudo apt-get install -y
           sudo apt-get remove -y llvm-12
           sudo apt-get install -y curl pkg-config findutils  wget
-          sudo apt install -y clang-11 cmake clang-format-11 clang-tidy-11
-          sudo apt-get install -y llvm-11 lldb-11 llvm-11-dev libllvm11 llvm-11-runtime
+          sudo apt install -y clang-13 cmake clang-format-13 clang-tidy-13
+          sudo apt-get install -y llvm-13 lldb-13 llvm-13-dev libllvm11 llvm-13-runtime
           sudo apt install -y python3 python3-pip
           sudo apt-get remove -y llvm-10
```
kuzminrobin, line 26: (nit)
I assume there are [some specific reasons](https://www.google.com/search?q=what%27s+the+difference+between+apt+and+apt-get&rlz=1C1GCEU_enUS858US858&oq=what%27s+the+difference+between+apt+and+&aqs=chrome.2.0i512l3j69i57j0i22i30l6.10121j1j7&sourceid=chrome&ie=UTF-8) for using sometimes `apt` and sometimes `apt-get` in L20 - L26 (rather than either `apt` or `apt-get` everywhere).

.github/workflows/trusted-ci-linux-on-main-pr.disabled
```diff
@@ -43,12 +43,12 @@ jobs:
...
           sudo apt-get install -y curl pkg-config findutils  wget
-          sudo apt install -y clang-11 cmake clang-format-11 clang-tidy-11 
-          sudo apt-get install -y llvm-11 lldb-11 llvm-11-dev libllvm11 llvm-11-runtime opt 
+          sudo apt install -y clang-13 cmake clang-format-13 clang-tidy-13 
+          sudo apt-get install -y llvm-13 lldb-13 llvm-13-dev libllvm11 llvm-13-runtime opt 
           sudo apt-get install -y nodejs npm
           sudo apt install -y python3 python3-pip 
           sudo apt-get remove -y llvm-10
-          sudo ln -s /usr/lib/llvm-11/bin/opt /usr/bin/opt
+          sudo ln -s /usr/lib/llvm-13/bin/opt /usr/bin/opt
```
kuzminrobin, line 51: (nit)
I have an impression that there is a chance of code duplication between multiple .yml files. If that's the case then consider extracting a common part into a separate .yml file and invoking the common part from multiple .yml files
(when and if you have time).

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:22.04
```
kuzminrobin, line 1: (nit)
I have an impression that 
`Docker/CI.Ubuntu20.dockerfile` has been renamed to
`Docker/CI.Ubuntu22.dockerfile` and some changes were made in the file content. 
But the result of those changes is shown such that the old file was completely removed and a brand new file has been created (the commit history is not preserved through the renaming). This typically happens when renaming is done with commands bypassing the git (rather than command `git mv old_path/old_name new_path/new_name`).

Just in case I'd like to make sure that you also considered the following alternatives.
A. If you want to preserve the commit history you can use git for renaming, 
`git mv Docker/CI.Ubuntu20.dockerfile Docker/CI.Ubuntu22.dockerfile`
then commit the change, then make changes to the file content, then commit the content change.

B. Renaming to 
`Docker/CI.Ubuntu.dockerfile`, such that the version of Ubunto is not in the file name, and you don't need to rename the file any more upon migration to newer Ubunto. You just make changes to the file itself.

C. Just adding 
`Docker/CI.Ubuntu22.dockerfile`, but not removing 
`Docker/CI.Ubuntu20.dockerfile`, such that the support for Ubuntu22 is added, but the support for Ubuntu20 is not removed.

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN apt-get install -y curl \
+  pkg-config \
+  findutils \
+  wget
+
+# build dependencies
+RUN apt-get install -y && \
+    apt-get remove -y llvm-12 clang-format-11 clang-tidy-11 && \
+    apt-get install -y curl pkg-config findutils  wget && \
+    apt install -y clang-13 cmake clang-format-13 clang-tidy-13 && \
```
kuzminrobin, line 17: (nit)
I assume that some future version of Ubunto is allowed to not contain the `llvm-12` or `clang-format-11` or `clang-tidy-11`, listed in L15, that is why the command "apt-get **remove**" (L15) can _fail_, which, combined with the `&&` operator in the end of the L15, can result in _skipping the commands in L16+_.

Consider executing the command in L15 _separately_. And _regardless of its success_ execute the commands in L14, L16+.

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+  apt-get install -y 
+
+RUN apt-get install -y curl \
+  pkg-config \
+  findutils \
+  wget
+
+# build dependencies
+RUN apt-get install -y && \
+    apt-get remove -y llvm-12 clang-format-11 clang-tidy-11 && \
```
kuzminrobin, line 15: What does the L14 install?

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+ADD ./requirements.txt /src/
+ADD ./manage /src/
+RUN cd /src/ && \
+      pip install -r requirements.txt && \
+      chmod +x manage
+
+# running the build
+ENV CC=clang-13 \
+  CXX=clang++-13 \
```
kuzminrobin, line 37: (nit) I assume that it is sufficient to have either L36-37 or L22-23, but hardly both.

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    npm install -g prettier@2.2.1 && \
+    export CC=clang-13 && \
+    export CXX=clang++ 
+
+# Python
+RUN apt install -y python3 python3-pip && \
+    update-alternatives --install /usr/bin/python python /usr/bin/python3 0
+
+ADD ./requirements.txt /src/
+ADD ./manage /src/
```
kuzminrobin, line 30: (nit, I don't know Docker) I assume that `ADD` _copies_ the files. What will be lost if the files are _moved_ instead?

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN cd /src/ && \
+      pip install -r requirements.txt && \
+      chmod +x manage
```
kuzminrobin, line 33: (nit, I don't know Docker) Is `manage` _our_ file or not? If it is ours, then why doesn't that file get born/created with the executable attribute from the very beginning (such that we don't need to set that attribute with the command `chmod +x manage`)?

qir/qat/Rules/Factory.cpp.orig
```diff
@@ -0,0 +1,801 @@
...
+                logger->internalError("Support for callable replacement is not implemented yet.");
+            }
+            else
+            {
+                throw std::runtime_error(
+                    "No logger present - internal error: Support for callable replacement is not implemented "
+                    "yet.\n");
+            }
+
+>>>>>>> main
```
kuzminrobin, line 163: L143 - 163 look like an unresolved conflict. This is strange for a newly created file. 
What is this file for? Did you forget to remove it?

qir/qat/Rules/Factory.cpp.orig
```diff
@@ -0,0 +1,801 @@
...
+            RuleSet::ReplaceDirection::ReplaceBackwards);
+        addRule(
+            {callByNameOnly("__quantum__rt__result_get_one"), deleteUnusedInstruction()},
+            RuleSet::ReplaceDirection::ReplaceBackwards);
+    }
+
+    void RuleFactory::resolveConstantArraySizes()
+    {
+        /// Array access replacement
+        auto size_replacer = [](Builder&, Value* val, Captures& cap, Replacements& replacements)
```
kuzminrobin, line 116: Did you really intend the L115 be a _doc_ comment? (`///` as opposed to `//`)
Multiple occurrences.

qir/qat/Rules/Factory.cpp.orig
kuzminrobin: Multiple occurrences.

qir/qat/Rules/Notation/Notation.cpp.orig
```diff
@@ -0,0 +1,75 @@
...
+                      ReplacementRule::Replacements& replacements) {
+                auto type  = val->getType();
+                auto instr = llvm::dyn_cast<llvm::Instruction>(val);
+
+                if (instr == nullptr)
+                {
+                    return false;
+                }
+
+>>>>>>> main
```
kuzminrobin, line 38: What is this file for? Forgot to remove? Otherwise there is an unresolved merge conflict.

qir/qsharp/Atomics/AdvancedConstantFolding/QSharpVersion/Example.csproj
```diff
@@ -1,8 +1,8 @@
-<Project Sdk="Microsoft.Quantum.Sdk/0.23.195983">
+<Project Sdk="Microsoft.Quantum.Sdk/0.24.201332">
```
kuzminrobin, line 1: (nit) Consider either removing the version (see below) or changing to `0.24.210930`.
```suggestion
<Project Sdk="Microsoft.Quantum.Sdk">
```

site-packages/tasks_ci/toolchain.py
```diff
@@ -48,12 +48,16 @@ def discover_tidy_impl(ext: str) -> str:
...
             return ret
 
     possible_commands = [
+        "/usr/bin/clang-tidy-13",
+        "/usr/local/bin/clang-tidy-13",
+        "/usr/bin/clang-tidy-12",
+        "/usr/local/bin/clang-tidy-12",
         "/usr/bin/clang-tidy-10",
         "/usr/bin/clang-tidy-11",
         "/usr/local/bin/clang-tidy-10",
```
kuzminrobin, line 63: A. In green L22 and L51 the version `12` is checked _last_ (even after `10`). But in L56 - 69 the version `12` is checked after `13`. It is suspicious.

B. In L61 - 64 the lines for version `10` (L61, L63) are interleaved with lines for version `11` (L62, L64). But in L57 - 60 the versions do not interleave. This is suspicious.

C. In L61 - 64 the version `10` is checked ahead of the version `11`. But in lines 22, 51 the version `11` is ahead of `10`. This is suspicious.

.github/workflows/ci-linux-on-main-pr.yml
```diff
@@ -26,14 +26,13 @@ jobs:
...
-          sudo apt-get remove -y llvm-12
+          sudo apt-get remove -y llvm-12 llvm-11 clang-format-11 clang-tidy-11 
           sudo apt-get install -y curl pkg-config findutils  wget
-          sudo apt install -y clang-11 cmake clang-format-11 clang-tidy-11 
-          sudo apt-get install -y llvm-11 lldb-11 llvm-11-dev libllvm11 llvm-11-runtime opt 
+          sudo apt install -y clang-13 cmake clang-format-13 clang-tidy-13 
+          sudo apt-get install -y llvm-13 lldb-13 llvm-13-dev libllvm11 llvm-13-runtime opt 
           sudo apt-get install -y nodejs npm
           sudo apt install -y python3 python3-pip 
-          sudo apt-get remove -y llvm-10
```
kuzminrobin, line 35: (nit) In some files the line `sudo apt-get remove -y llvm-10` is removed, but in the others it is not. It is suspicious.

qir/qat/Rules/Factory.cpp.orig
kuzminrobin: Multiple `.orig` files.

kuzminrobin (approved):

.github/workflows/ci-linux-on-main-pr.yml
troelsfr: I agree that it could make sense to use a environment variable if upgrades would happen often. The downside is that it would lower readability and I think that given that we do not expect to upgrade LLVM more than once a year at most, I would prefer the to keep it the way it is.

.github/workflows/trusted-ci-linux-on-main-pr.disabled
troelsfr: This is a good suggestion - I was not aware that you could do this with Github actions. I will do this change once `qiskit-qir` is ready to be reinstated in the test pipelines.

Docker/CI.Ubuntu22.dockerfile
troelsfr: As such you are right that it is a rename and the appropriate way to have done that would be to use `git mv`. In this case, we are removing support for Ubuntu 20 as there is no straight forward path to install LLVM 13 on Ubuntu 20. I will keep `git mv` in mind for next time.

Docker/CI.Ubuntu22.dockerfile
troelsfr: When building a docker image, it is common practice to concatenating lines using `&&`. The reason for this is that every command `RUN`/`ADD`/etc. creates a new cache that potentially takes up a lot of space on your hard drive. I agree that this ruins debuggability, but unfortunately not doing this will exhaust the hard drive quickly.

Docker/CI.Ubuntu22.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN apt-get install -y && \
+    apt-get remove -y llvm-12 clang-format-11 clang-tidy-11 && \
+    apt-get install -y curl pkg-config findutils  wget && \
+    apt install -y clang-13 cmake clang-format-13 clang-tidy-13 && \
+    apt-get install -y llvm-13 lldb-13 llvm-13-dev libllvm11 llvm-13-runtime opt && \
+    apt-get install -y nodejs npm && \
+    apt install -y python3 python3-pip && \
+    npm install -g prettier@2.2.1 && \
+    export CC=clang-13 && \
+    export CXX=clang++
```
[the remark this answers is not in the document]
troelsfr: Fixed.

Docker/CI.Ubuntu22.dockerfile
troelsfr: It is redundant - it has been removed. Thanks!

Docker/CI.Ubuntu22.dockerfile
troelsfr: You are right.

Docker/CI.Ubuntu22.dockerfile
troelsfr: The command `ADD` copies a file from the host machine to the docker image ("VM" image). If this would be replaced with a move instruction, the data on the host machine would vanish which means that you'd not be able to run the build process twice in a row. I am not aware that there is an alternative to `ADD` but even if there was, I would hesitate using it because of the previous explanation.

Docker/CI.Ubuntu22.dockerfile
troelsfr: `manage` is our own tool which we copy into the image. In principle, we do not need `+x` as the file should already have that. This is merely a precaution to ensure that it has not lost its executable status for any reason.

qir/qat/Rules/Factory.cpp.orig
troelsfr: All removed. Thanks!

qir/qat/Rules/Notation/Notation.cpp.orig
troelsfr: Yes, it is trash.

qir/qsharp/Atomics/AdvancedConstantFolding/QSharpVersion/Example.csproj
troelsfr: All versions was updated on main and latest merge should address this.

site-packages/tasks_ci/toolchain.py
troelsfr: Version ordering fixed.

.github/workflows/ci-linux-on-main-pr.yml
troelsfr: Thanks, this has been updated.

site-packages/tasks_ci/toolchain.py
kuzminrobin: FYI: In L22 the order is still different (`12` is the last, after `10`).

.github/workflows/ci-linux-on-main-pr.yml
kuzminrobin: FYI: Still in `.github/workflows/trusted-ci-linux-on-main-pr.disabled` L50.

kuzminrobin:
Re-approved.

.github/workflows/ci-linux-on-main-pr.yml
```diff
@@ -26,15 +26,15 @@ jobs:
         run: |
           sudo apt-get update -y
           sudo apt-get install -y 
-          sudo apt-get remove -y llvm-12 llvm-11 clang-11 clang-format-11 clang-tidy-11 clang-format-14 clang-tidy-14 
+          sudo apt-get remove -y llvm-14 llvm-12 llvm-11 clang-11 clang-format-11 clang-tidy-11 clang-format-14 clang-tidy-14
```
kuzminrobin, line 29: Consider grouping the versions together:
```suggestion
          sudo apt-get remove -y llvm-14 clang-format-14 clang-tidy-14 llvm-12 llvm-11 clang-11 clang-format-11 clang-tidy-11
```

.github/workflows/ci-mac-on-main-pr.yml
```diff
@@ -101,6 +106,39 @@ jobs:
...
+          export QAT_BINARY=${PWD}/Debug/qir/qat/Apps/qat     
+
+          pushd qir/qiskit-tests/
+          pip3 install pytest
+
+          python3 -m pytest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CC: clang
+          CXX: clang++
```
kuzminrobin, line 140: As I understand the `env:` section sets the environment first, and then the `run:` section executes the commands.
I'm curious, what is the advantage of placing the `env:` section _after_ the `run:`?

.github/workflows/target-testing.yml
```diff
@@ -26,16 +26,15 @@ jobs:
         run: |
           sudo apt-get update -y
           sudo apt-get install -y 
-          sudo apt-get remove -y llvm-12
+          sudo apt-get remove -y llvm-12 llvm-14 llvm-11 clang-11 clang-format-11 clang-tidy-11 clang-format-14 clang-tidy-14
```
kuzminrobin, line 29: FYI: Consider re-ordering by version (one more case) below. You also probably need to remove `clang-12`, `clang-format-12`? 
```suggestion
          sudo apt-get remove -y llvm-14 clang-format-14 clang-tidy-14 llvm-12 llvm-11 clang-11 clang-format-11 clang-tidy-11 
```

.github/workflows/ci-mac-on-main-pr.yml
```diff
@@ -101,6 +106,39 @@ jobs:
...
+
+          git clone https://github.com/microsoft/qiskit-qir.git
+          pushd qiskit-qir
+
+          pip3 install -e .
+          popd
+
+          echo "Testing installation"
+          python -c "from qiskit_qir import to_qir"
+          echo "Done"
```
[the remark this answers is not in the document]
troelsfr: The CI would fail and we would know that it is because `qiskit-qir` was not correctly installed.

qir/qat/RemoveDisallowedAttributesPass/RemoveDisallowedAttributesPass.hpp
```diff
@@ -21,20 +21,28 @@ namespace quantum
         RemoveDisallowedAttributesPass()
           : allowed_attrs_{
                 static_cast<String>("EntryPoint"), static_cast<String>("InteropFriendly"),
-                static_cast<String>("requiredQubits"), static_cast<String>("requiredResults")}
+                static_cast<String>("requiredQubits"), static_cast<String>("requiredResults")
```
kuzminrobin, line 24: If the `String` here is an alias of `std::string` then what is the advantage of 
`static_cast<String>("requiredQubits")` over the 
`String("requiredQubits")`?

.github/workflows/ci-mac-on-main-pr.yml
troelsfr: Your understanding is correct. There is no difference depending on order - the section acts as a dictionary and the sections are used for their respective tasks (i.e. `env` is looked up to set environment, `run` to execute the script)

qir/qat/RemoveDisallowedAttributesPass/RemoveDisallowedAttributesPass.hpp
troelsfr: As far as I am aware there is no difference as `static_cast<String>("EntryPoint")` should translate into `String("EntryPoint")`. It is only a semantic difference in that we are casting `char const *` to `String`, rather than creating `String`. In the case of`std::string` I believe it is even possible to get away with implicit casts (but I might remember wrong, but I don't think the constructor is marked as explicit).

kuzminrobin:
Re-approved.

kuzminrobin (approved):

PR #90, opened 2022-06-01
Fixing nonnull survival

troelsfr:
This PR adds a check to remove `nonnull` when qubit indices are reindexed.


qir/qat/StaticResourceComponent/QubitRemapPass.cpp
```diff
@@ -84,7 +84,7 @@ namespace quantum
                 }
                 break;
             case AllocationAnalysis::ResultResource:
-                if (results_mapping.find(index) == results_mapping.end())
+                if (results_mapping.find(index) != results_mapping.end())
```
bettinaheim, line 87: Why isn't there a unit test that caught this? It would be good to add a test - that seems like a gap.

bettinaheim: Answered offline: The Result handling here is set up following the same pattern as for Qubits, except we don't have a Reset for Results, and hence no tests.

bettinaheim: Answered offline: The Result handling here is set up following the same pattern as for Qubits, except we don't have a Reset for Results, and hence no tests.

bettinaheim (approved):

PR #94, opened 2022-06-02
Validation for qubits and result usage

troelsfr:
This PR introduces tests of whether qubits and results are being used. It also centralises error messages and makes testable.


qir/qat/Validator/Tests/Unit/RequiredQubitsAndResults.cpp
```diff
@@ -0,0 +1,153 @@
...
+    EXPECT_TRUE(logger->raiseNoQubitsError());
+    EXPECT_FALSE(logger->raiseNoResultsError());
+}
+
+TEST(FunctionValidationPass, NothingMissing)
+{
+    auto logger = std::make_shared<MockLogger>();
+    testSkeleton(
+        R"script(  
+  tail call void @__quantum__reset__result(%Result* nonnull inttoptr (i64 2 to %Result*))
```
idavis, line 147: What is `__quantum__reset__result`? I can't say I've seen that before and it isn't `qis` or `rt` prefixed. Is it just an extern function to force result allocation in the test, and then you don't run the extern function check?

qir/qat/ValidationPass/FunctionValidationPass.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "Llvm/Llvm.hpp"
+
+#include <functional>
+#include <unordered_map>
+#include <vector>
+
+namespace microsoft
+{
+namespace quantum
+{
```
kuzminrobin, line 19: Doesn't the QAT compiler and chosen C++ standard support this?
```suggestion
namespace microsoft::quantum
{
```

qir/qat/ValidationPass/FunctionValidationPass.hpp
```diff
@@ -0,0 +1,63 @@
...
+        using ILoggerPtr  = ILogger::ILoggerPtr;
+        using Location    = ILogger::Location;
+        using StringRef   = llvm::StringRef;
+
+        enum ResourceType
+        {
+            None,
+            Qubit,
+            Result
+        };
```
kuzminrobin, line 35: (nit) It is somewhat surprising to see an `enum` added, but not used in the _newly added_ code. I assume this enum is planned to be used in the future.

qir/qat/ValidationPass/FunctionValidationPass.cpp
```diff
@@ -0,0 +1,49 @@
...
+            logger_->errorNoResultsPresent(&function);
+        }
+
+        return llvm::PreservedAnalyses::all();
+    }
+
+    bool FunctionValidationPass::isRequired()
+    {
+        return true;
+    }
```
kuzminrobin, line 47: (personal curiosity)
It is not the first PR of yours where I see this kind of _simple_ implementation for a _static_ member func. Before your PRs I've mostly seen either a more complex implementation (of a static member func) or a non-static member func (typically a virtual func override) with similar simple implementation.
Is the approach that you use a paradigm/template? If yes, then can you please refer me to that template?
What if the instance of this class is manipulated with the base class pointer, and this static member function is called? (I assume the base class also has this member function and it is also static) The bases class' version of this function will be invoked, right? If yes, then what such an approach is intended for, and why is this approach implemented with the _static_ member functions?

kuzminrobin (approved):

qir/qat/Validator/Tests/Unit/RequiredQubitsAndResults.cpp
troelsfr: It is just a made up function for the purpose of testing since I did not find a function that only takes a `Result*` as argument in the spec.

qir/qat/Logging/ILogger.cpp
```diff
@@ -62,5 +62,116 @@ namespace quantum
...
+    }
+
+    void ILogger::errorExpectedStraightLineCodeMultipleFunctions(llvm::Value* ptr)
+    {
+        if (ptr)
+        {
+            setLocationFromValue(ptr);
+        }
+        error("Expected straight line code, but multiple functions present.");
+    }
```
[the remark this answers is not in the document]
troelsfr: This is a good suggestion - will revise.

qir/qat/Validator/Tests/Unit/RequiredQubitsAndResults.cpp
troelsfr: I've changed the name to avoid any confusion.

qir/qat/ValidationPass/FunctionValidationPass.hpp
troelsfr: It is a mistake that it is there.

qir/qat/ValidationPass/FunctionValidationPass.cpp
troelsfr: This is a LLVM function that is mandatory which tells the pass managers whether the pass may be skipped or not. It is used in [PassInstrumentation](https://llvm.org/doxygen/PassInstrumentation_8h_source.html#l00217). I have not read about it in detail and it may be that you can skip adding this function. However, it is unclear what the default behaviour would be without reading the LLVM source in detail.

qir/qat/ValidationPass/FunctionValidationPass.cpp
troelsfr: I wasn't aware that C++14 has support for this. It is a good suggestion and I would propose to make a separate PR where this is changed repository wide and not just in a single file.

kuzminrobin: I'm leaving this decision to you 
whether to make it in a separate PR repo-wide 
or in the new files and then at some time in a separate PR repo-wide
(or other or not do :-) ).

PR #96, opened 2022-06-14
Feature/poison value check

troelsfr:


qir/qat/ValidationPass/ValidationPassConfiguration.hpp
```diff
@@ -100,6 +103,9 @@ class ValidationPassConfiguration
 
     bool requires_qubits_{false};
     bool requires_results_{false};
+
+    bool allow_poison_{true};
+    bool allow_undef_{true};
```
swernli, line 108: Would it make sense to disallow these values by default? Not a big deal right now, as we can supply the necessary flags when calling and eventually we'll have pass configurations associated with the profiles themselves.

swernli (approved):

qir/qat/ValidationPass/ValidationPassConfiguration.hpp
troelsfr: We need to go over all settings and decide what is default and what is not. Also, we should make settings configure by the profile

PR #103, opened 2022-06-21
Add lower switch pass in static resource

swernli:
Since the static resource pass also performs a SimplifyCFG pass, it can have have the side effect of reintroducing switch instructions even after they've been removed by the earlier LowerSwitch pass. This adds LowerSwitch to the set of post id change passes.


troelsfr (approved):

troelsfr:
I think it would have been great to also have a test for this scenario, but other than that, the PR is good.

PR #107, opened 2022-06-29
Setting error flag on division by zero

troelsfr:
This PR introduces a global error variable which is set if the program execution experience a division by zero. For instance,
```llvm
entry:
  %y = alloca i64, align 8
  %x = alloca i64, align 8
  store i64 0, i64* %x, align 4
  store i64 20, i64* %y, align 4
  %0 = load i64, i64* %x, align 4
  %1 = load i64, i64* %y, align 4
  %2 = sdiv i64 %0, %1
  call void @print_i64(i64 %2)
```
is transformed into
```llvm
entry:
  %y = alloca i64, align 8
  %x = alloca i64, align 8
  store i64 0, i64* %x, align 4
  store i64 20, i64* %y, align 4
  %0 = load i64, i64* %x, align 4
  %1 = load i64, i64* %y, align 4
  %2 = icmp eq i64 %1, 0
  br i1 %2, label %if_denominator_is_zero, label %after_zero_check

if_denominator_is_zero:                           ; preds = %entry
  %3 = load i64, i64* @__qir__error_code, align 4
  %4 = icmp eq i64 %3, 0
  br i1 %4, label %if_ecc_not_set, label %ecc_set_finally

if_ecc_not_set:                                   ; preds = %if_denominator_is_zero
  store i64 1338, i64* @__qir__error_code, align 4
  br label %ecc_set_finally

ecc_set_finally:                                  ; preds = %if_denominator_is_zero, %if_ecc_not_set
  br label %after_zero_check

after_zero_check:                                 ; preds = %entry, %ecc_set_finally
  %5 = sdiv i64 %0, %1
  call void @print_i64(i64 %5)
```


qir/qat/DivisionByZeroPass/DivisionByZeroPass.cpp
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "qir/qat/DivisionByZeroPass/DivisionByZeroPass.hpp"
```
kuzminrobin, line 4: If you need to change the names of the directories or move the files to different dirs, then you'll have to change the source code (this line).
Consider minimizing the use of the paths in the source code. Consider passing the paths (to the headers) as the command-line args to the compiler (`-Iqir/qat/DivisionByZeroPass`) and not using the paths in the source:
```suggestion
#include "DivisionByZeroPass.hpp"
```

qir/qat/DivisionByZeroPass/DivisionByZeroPass.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "qir/qat/QatTypes/QatTypes.hpp"
+
+#include <functional>
+#include <unordered_map>
+#include <vector>
+
+namespace microsoft::quantum
+{
+char* const   DivisionByZeroPass::EC_REPORT_FUNCTION      = "__qir__report_error_value";
+char* const   DivisionByZeroPass::EC_VARIABLE_NAME        = "__qir__error_code";
```
kuzminrobin, line 17: Currently it is const pointer to non-const data (while the data is the string literal residing in the read-only data segment).
Consider disabling the update of the string literal:
```suggestion
const char* const   DivisionByZeroPass::EC_REPORT_FUNCTION      = "__qir__report_error_value";
const char* const   DivisionByZeroPass::EC_VARIABLE_NAME        = "__qir__error_code";
```

qir/qat/DivisionByZeroPass/DivisionByZeroPass.cpp
```diff
@@ -0,0 +1,152 @@
...
+            }
+            break;
+        }
+    }
+
+    if (entry)
+    {
+        for (auto start_block : exit_blocks)
+        {
+            auto if_block    = start_block->splitBasicBlock(start_block->getTerminator(), "if_error_occured", false);
```
kuzminrobin, line 89: I assume that the GitHub warning for this line means that it should be:
```suggestion
            auto if_block    = start_block->splitBasicBlock(start_block->getTerminator(), "if_error_occurred", false);
```

qir/qat/TestTools/TestVm.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+    template <typename R, typename... Args> R run(String const& name, Args&&... args);
+    template <typename R> R                   run(String const& name);
+
+  private:
+    void buildVm()
+    {
+        llvm::ExitOnError exit_on_error;
+
+        // TODO(unknown): remove exit_on_error
```
kuzminrobin, line 85: What does the `unknown` mean? 
It is unknown who has written this TODO? Or it is unknown which bug/task this TODO relates to?
(more than one occurrence)

qir/qat/TestTools/Tests/Unit/vm.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "gtest/gtest.h"
+#include "qir/qat/TestTools/TestVm.hpp"
+
+using namespace microsoft::quantum;
+
+int8_t print_call_value = 0;
```
kuzminrobin, line 9: (nit)
The use of a global var. (Theoretically) If multiple tests use it, and they are executed in parallel (in different threads), the behavior can be unexpected.
Consider using a local var instead and passing a pointer or a ref to it as an arg to the funcs that use it, and capturing it by ref in the lambdas that use it.

kuzminrobin (approved):

qir/qat/TestTools/Tests/Unit/vm.cpp
troelsfr: I agree that use of global variables should be discouraged. However, inside a test I think it does very little harm and  using a local or arg does not work in this case.. Re. thread-safety, making the variable local would not solve the issue. Instead you would need to either make it atomic or use  a mutex to lock it. I think the real danger is linking errors due to replicated symbols across translation units. To circumvent this, I've added an anonymous namespace around it.

qir/qat/TestTools/TestVm.hpp
troelsfr: `unknown` is added by the linter when a username or issue is not present. These TODOs are deprecated anyway, so they should just go.

qir/qat/DivisionByZeroPass/DivisionByZeroPass.cpp
troelsfr: There are both pros and cons to using repo absolute paths. The downside is, as you mention, that refactoring becomes slightly harder. The upside is that it removes ambiguity around which file is used. I.e. if two passes both have a file that is called `Config.hpp` then doing `-Iqir/qat/Pass1 -Iqir/qat/Pass2` will confuse which file is being included and you would consequently need longer filenames to resolve the matter. For this reason, I feel reluctant to make this change.

PR #110, opened 2022-07-12
Record pull back

troelsfr:
This PR pulls all function calls ending with `_record_output` back to the end of the block.


qir/qat/Generators/ProfileGenerator.cpp
```diff
@@ -277,6 +278,14 @@ void ProfileGenerator::setupDefaultComponentPipeline()
             {
                 fpm.addPass(ZExtTransformPass());
             }
+
+            llvm::errs() << "---- xxxxx ----\n";
```
bettinaheim, line 282: Should this be here?

qir/qat/RecordPullBackPass/RecordPullBackPass.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <stdexcept>
+#include <unordered_map>
+#include <vector>
+
+namespace microsoft::quantum
+{
+
+std::string const RecordPullBackPass::RECORD_INSTR_END = "_record_output";
+
+llvm::PreservedAnalyses RecordPullBackPass::run(llvm::Function& function, llvm::FunctionAnalysisManager& /*mam*/)
```
bettinaheim, line 22: There are two issues with this pass: 

1) The immediate need is to ensure that all quantum instructions that measurements are followed only be other measurements, record output functions (and the ret void), when the initial IR mixes measurements and other __quantum__qis* calls. You can rely on only record output functions depending on measurement results and nothing else. 

2) Not relevant to solve the immediate need 1, but general feed back on this logic: I believe we still have an issue with the logic separation not recognizing record output function. Running this pass after logic separation does not address the need since the logic separation distroys the order of the record output functions.

qir/qsharp/TargetPackages/QSharpVersion/Example.qs
```diff
@@ -1,26 +1,16 @@
-namespace Testing {
```
bettinaheim, line 1: I am not sure why this example was changed (or why a Q# example is checked in in the first place), but I assume that it is not used for testing.

troelsfr: This was the example you've sent me which was accidentally checked into the earlier PR which made it to main. This PR just changes it back to whee it was. The `TargetPackages` directory is just an example project that shows how to configure Q# to the different targets. It is not used in testing.

PR #111, opened 2022-07-14
Stricter compile flags

troelsfr:
This PR sets stricter compile flags for our own library to ensure accidental mistakes happen.


.bazelrc
```diff
@@ -22,14 +22,6 @@ build:generic_clang --repo_env=CC=clang
 build:generic_clang --cxxopt=-std=c++17
```
kuzminrobin, line 22: Above, in L4, consider teh GitHub warning for spelling of the word "unfortunat**t**ely"

config/microsoft.bzl
```diff
@@ -0,0 +1,58 @@
+"""
+Target definitions to enable stricter C++ stanard for internal libraries
```
kuzminrobin, line 2:
```suggestion
Target definitions to enable stricter C++ standard for internal libraries
```

kuzminrobin (approved):

.bazelrc
troelsfr: Fixed.

PR #112, opened 2022-07-15
Adding profile configuration to the profile flag

troelsfr:
This PR sets configuration according to the profile selected. For instance, with the qat version in this PR, one can  use the command:
```
qat --profile provider_b340 ...
``` 
instead of
```
qat --apply --verify-module --validate --disable-grouping --always-inline --reindex-qubits --requires-qubits --requires-results --no-disable-record-output-support --unroll-loops --replace-qubit-on-reset --pull-records-back --entry-point-attr EntryPoint --profile provider_b340 ...
```


qir/qat/Apps/Qat/ProfileConfiguration.hpp
```diff
@@ -0,0 +1,305 @@
...
+    configuration_manager.updateParameter("disable-reference-counting", true);
+    configuration_manager.updateParameter("disable-alias-counting", true);
+    configuration_manager.updateParameter("disable-string-support", true);
+    configuration_manager.updateParameter("disable-record-output-support", false);
+
+    configuration_manager.updateParameter("allow-internal-calls", false);
+    configuration_manager.updateParameter("requires-qubits", true);
+    configuration_manager.updateParameter("requires-results", true);
+    configuration_manager.updateParameter("allow-poison", true);
+    configuration_manager.updateParameter("allow-undef", true);
```
swernli, line 230: It looks like this chunk got duplicated below for this profile.

qir/qat/Apps/Qat/ProfileConfiguration.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+    configuration_manager.updateParameter("annotate-qubit-use", true);
+    configuration_manager.updateParameter("annotate-result-use", true);
+    configuration_manager.updateParameter("annotate-max-qubit-index", true);
+    configuration_manager.updateParameter("annotate-max-result-index", true);
+    configuration_manager.updateParameter("reindex-qubits", true);
+    configuration_manager.updateParameter("replace-qubit-on-reset", false);
+    configuration_manager.updateParameter("inline-after-id-change", true);
+
+    configuration_manager.updateParameter("separate-circuits", true);
```
swernli, line 101: I'm a little surprised that this experimental setting is On by default (looks like it has been for a while). What effect does this have? Is this ignored when grouping is overall disabled (as on line 39)?

troelsfr: Yes, you are right. The default should be off. I believe it has had no effect since production is always ran with `--disable-grouping` which effectively put it off.

qir/qat/Apps/Qat/ProfileConfiguration.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+    configuration_manager.updateParameter("allow-internal-calls", false);
+    configuration_manager.updateParameter("requires-qubits", true);
+    configuration_manager.updateParameter("requires-results", true);
+    configuration_manager.updateParameter("allow-poison", true);
+    configuration_manager.updateParameter("allow-undef", true);
```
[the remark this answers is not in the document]
troelsfr: Yes, I agree.

qir/qat/Apps/Qat/ProfileConfiguration.hpp
troelsfr: Yes, this needs to be fixed.

qir/cli-tests/BUILD
```diff
@@ -0,0 +1,15 @@
...
+        "test_profile_cli_config.py",
+    ],
+    args = [
+        "$(location //qir/qat:qat)",
+    ],
+    data = [
+        "//qir/qat",
+    ],
+    main = "test_profile_cli_config.py",
+)
```
kuzminrobin, line 15: It is undesirable that 
* there is a dash-separated (`test-profile-cli-config`) and underscore-separated (`test_profile_cli_config`) identifier for the same thing. This encourages a confusion, and sometimes the bugs hard to catch. Consider using the same identifier for all cases;
* the same string literal ("test_profile_cli_config.py") occurs more than once, consider de-duplicating.

qir/cli-tests/test_profile_cli_config.py
```diff
@@ -0,0 +1,114 @@
+import subprocess
+import sys
+
+TESTS = {
+    "provider_4bf9": [
```
kuzminrobin, line 5: May be it is possible to use a more sensible fragment instead of `4bf9`?
Otherwise how can the reader/user memorize it and make sure that `4bf9` is the right fragment to use here?

qir/cli-tests/test_profile_cli_config.py
```diff
@@ -0,0 +1,114 @@
...
+            print("Testing " + profile + " " + "." * (20 - len(profile)) + "[  OK  ]")
+        else:
+            fail = True
+            print("Testing " + profile + " " + "." * (20 - len(profile)) + "[ FAIL ]")
+            lines1 = out1.split("\n")
+            lines2 = out2.split("\n")
+
+            for i in range(len(lines1)):
+                l1 = lines1[i]
+                l2 = lines2[i]
```
kuzminrobin, line 107: I have a suspicion that the fragment `lines2[i]` can access outside of `lines2`, at least I don't see any guarantee that prevents that.

qir/cli-tests/test_profile_cli_config.py
```diff
@@ -0,0 +1,114 @@
...
+
+                if l1 != l2:
+                    print("- {}".format(l1))
+                    print("+ {}".format(l2))
```
kuzminrobin, line 111: Consider renaming the identifiers. In some editors/fonts the identifiers `l2` looks like the number `12`.

qir/cli-tests/test_profile_cli_config.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if out1 == out2:
+            print("Testing " + profile + " " + "." * (20 - len(profile)) + "[  OK  ]")
+        else:
+            fail = True
+            print("Testing " + profile + " " + "." * (20 - len(profile)) + "[ FAIL ]")
+            lines1 = out1.split("\n")
+            lines2 = out2.split("\n")
+
+            for i in range(len(lines1)):
```
troelsfr, line 105:
```suggestion
            for i in range(min(len(lines1), len(lines2)):
```

qir/qat/Apps/Qat/ProfileConfiguration.hpp
```diff
@@ -0,0 +1,292 @@
...
+    configuration_manager.updateParameter("allow-peeling", true);
+    configuration_manager.updateParameter("allow-runtime", true);
+    configuration_manager.updateParameter("allow-upper-bound", true);
+    configuration_manager.updateParameter("allow-profile-based-peeling", true);
+    configuration_manager.updateParameter<uint64_t>("full-unroll-count", 1024);
+    configuration_manager.updateParameter<int32_t>("unroll-opt-level", 3);
+    configuration_manager.updateParameter("only-when-forced", false);
+    configuration_manager.updateParameter("forget-scev", false);
+    configuration_manager.updateParameter("always-inline", true);
+    configuration_manager.updateParameter<int32_t>("inlining-parameter", 2147483647);
```
kuzminrobin, line 74: How can the reader/user/maintainer make sure that the "_magic number_" `2147483647` is correct? 
May be a hexadecimal representation can make more sense?

qir/cli-tests/test_profile_cli_config.py
troelsfr: You are right. For loop has been patched.

qir/qat/Apps/Qat/ProfileConfiguration.hpp
```diff
@@ -0,0 +1,292 @@
...
+    configuration_manager.updateParameter("separate-circuits", false);
+
+    configuration_manager.updateParameter("allow-internal-calls", false);
+    configuration_manager.updateParameter("requires-qubits", true);
+    configuration_manager.updateParameter("requires-results", true);
+    configuration_manager.updateParameter("allow-poison", false);
+    configuration_manager.updateParameter("allow-undef", false);
+}
+
+inline void configureProviderb340Profile(ConfigurationManager& configuration_manager)
```
kuzminrobin, line 186: It is not something that I typically see when such a large function 
* is defined in a header rather than a .cpp file; each .cpp file that includes this header may get a separate copy of this function body during compilation, and those copies may stay not merged into a single copy during linking;
* is recommended to be inlined (`inline`).

qir/cli-tests/test_profile_cli_config.py
troelsfr: Updated.

qir/cli-tests/BUILD
kuzminrobin: OK.

kuzminrobin (approved):

qir/qat/Apps/Qat/ProfileConfiguration.hpp
troelsfr: Done.

qir/qat/Apps/Qat/ProfileConfiguration.hpp
troelsfr: Fixed.

PR #123, opened 2022-07-25
Updating profile for provider `provider_7ee0`

troelsfr:


qir/qat/ValidationPass/ValidationPassConfiguration.cpp
```diff
@@ -177,6 +177,7 @@ ValidationPassConfiguration ValidationPassConfiguration::fromProfileName(String
             "__quantum__rt__tuple_end_record_output:void ()",
             "__quantum__rt__array_start_record_output:void ()",
             "__quantum__rt__array_end_record_output:void ()",
+            "__quantum__rt__bool_record_output:void (i1, i8*)",
```
cesarzc, line 180:
```suggestion
            "__quantum__rt__result_record_output:void (%Result*, i8*)",
            "__quantum__rt__bool_record_output:void (i1, i8*)",
            "__quantum__rt__int_record_output:void (i64, i8*)",
```

cesarzc: We would also need to add the output record functions for result and int.

cesarzc (approved):

PR #124, opened 2022-07-25
Refactoring record pulling

troelsfr:


qir/cli-tests/test_profile_cli_config.py
```diff
@@ -57,7 +57,7 @@
         "--no-disable-record-output-support",
         "--unroll-loops",
         "--replace-qubit-on-reset",
-        "--pull-records-back",
+        "--use-one-shot-measurements",
```
swernli, line 60: Unfortunately, the term "shot" is heavily associated with another meaning that might cause confusion here (see for example how it is used in the az quantum CLI parameters here: https://docs.microsoft.com/en-us/cli/azure/quantum/job?view=azure-cli-latest#az-quantum-job-submit-optional-parameters). Since it usually refers to separate executions of the same program, a different term should be used here.  Maybe something like `--defer-measurements` to connect it logically with the [Deferred Measurement Principle](https://en.wikipedia.org/wiki/Deferred_Measurement_Principle)?

troelsfr: OK, will refactor.

PR #126, opened 2022-08-01
Add phi instruction to provider_7ee0

cesarzc:
Add `phi` instruction to `provider_7ee0` to allow programs that use branching.


kuzminrobin (approved):

troelsfr (approved):
It is confirmed that `phi` nodes are supported for this profile.

PR #129, opened 2022-08-09
Improved compiler support & clean up

troelsfr:
The overall purpose of this PR is to prepare for cross-compilation by 1) creating stricter compile (and hence lower the risk of compile failure with other compilers than clang) and 2) adding experimental support for cross compilation from macOS (`Aarch64` or `x86`) til Linux `x86`.

This PR further cleans the repository for unused code. Note that the mentioned docker images needs to be commented ut in the `BUILID` files to work. These are deactivated as Docker causes issues on the Windows CI.

# Stricter compile flags & code clean up
This PR introduces stricter compile flags for Bazel. Along with this, it cleans a the code a number of places to accommodate the stricter flags.

# Updated Markdown configuration

The markdown configuration is updated to follow that of the organisation.

# Experimental cross-compilation of Docker images

This repository now offers support for building Docker images using `bazel`:
```sh
bazelisk run //qir/qat:qat-image --config generic_clang --config release   --platforms=@//:linux-x86
```
This (cross-)compiles `qat` and creates a Docker image containing the executable using `x86_64` architecture irrespective of whether the host platform is `aarch64`. Running this command on macOS M2, the corresponding Docker image is:
```
 % docker image inspect bazel/qir/qat:qat-image | grep Architecture
        "Architecture": "amd64
% docker run --rm -it --platform linux/amd64 bazel/qir/qat:qat-image /opt/qat-x86
Usage: /opt/qat-x86 [options] filename
```

# Experimental support Linux cross compiling for OS X
To cross compile on Apple silicon, start by installing GCC for Linux:
```
brew tap messense/macos-cross-toolchains
brew install x86_64-unknown-linux-gnu
```
To create Linux binaries specify the platform `--platforms=@//:linux-x86`. On macOS M2, this the process looks like:
```
% bazelisk build //qir/qat:qat --config generic_clang --config release   --platforms=@//:linux-x86
INFO: Analyzed target //qir/qat:qat (2 packages loaded, 214 targets configured).
INFO: Found 1 target...
Target //qir/qat:qat up-to-date:
  bazel-bin/qir/qat/qat
INFO: Elapsed time: 8.495s, Critical Path: 8.19s
INFO: 7 processes: 1 internal, 6 darwin-sandbox.
INFO: Build completed successfully, 7 total actions
```
resulting in following executable:
```
% file bazel-bin/qir/qat/qat
bazel-bin/qir/qat/qat: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.16, not stripped
```
This PR also adds target platforms `apple-x86` and `apple-silicon`. Compiling for `x86` on `aarch64`:
```
bazelisk build //qir/qat:qat --config generic_clang --config release   --platforms=@//:apple-x86
INFO: Analyzed target //qir/qat:qat (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //qir/qat:qat up-to-date:
  bazel-bin/qir/qat/qat
INFO: Elapsed time: 0.835s, Critical Path: 0.21s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```
which produces an `x86` binary for Intel based Apple machines:
```
 % file bazel-bin/qir/qat/qat
bazel-bin/qir/qat/qat: Mach-O 64-bit executable x86_64
```
And vice versa [host: `x86`], 
```
% bazelisk build //qir/qat:qat --config generic_clang --config release   --platforms=@//:apple-silicon
INFO: Analyzed target //qir/qat:qat (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //qir/qat:qat up-to-date:
  bazel-bin/qir/qat/qat
INFO: Elapsed time: 1.946s, Critical Path: 0.84s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```
yields
```
% file bazel-bin/qir/qat/qat
bazel-bin/qir/qat/qat: Mach-O 64-bit executable arm64
```


qir/qat/BUILD
```diff
@@ -423,3 +439,23 @@ ms_cc_binary(
...
+
+pkg_tar(
+    name = "qat-tar",
+    srcs = [":qat"],
+    mode = "0755",
+    package_dir = "/opt",
+    visibility = ["//visibility:public"],
+)
+
+# To enable building Docker images, add this
```
swernli, line 451: Just to clarify for my own understanding, this means a developer would need to modify this file to uncomment these lines before they could use the image build option, correct? Does leaving this on by default cause problems if the environment doesn't have docker, for example?

qir/qat/TestTools/Tests/Unit/BUILD
```diff
@@ -1,27 +1,25 @@
...
-        "//qir/qat:QatTypes",
-        "//qir/qat:Rules",
-        "//qir/qat:StaticResourceComponent",
-        "//qir/qat:TestTools",
-        "//qir/qat:TransformationRulesPass",
-        "//qir/qat:ValidationPass",
-        "//qir/qat:Validator",
-    ],
-)
+# cc_test(
```
swernli, line 1: Why keep this file if it's fully commented out? Is it going to be used later?

WORKSPACE
```diff
@@ -76,3 +101,31 @@ http_archive(
...
+    container_repositories = "repositories",
+)
+
+container_repositories()
+
+load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")
+
+container_deps()
+
+# To enable building Docker images, add this
```
swernli, line 125: Looks like this has to be uncommented as well to enable container building. When this does get promoted from experimental to full feature, it would be great to have this logic behind a single flag to make opt-in/out easy.

swernli (approved):
A couple minor comments, but otherwise looks good!

qir/qat/BUILD
troelsfr: Yes, that is correct. These lines need uncommenting. The reason is that docker builds does not work on native Windows PowerShell and would hence break CI. I have not figured out a way enable this target without it affecting the Windows builds, but it is something I am looking into (low priority though).

qir/qat/TestTools/Tests/Unit/BUILD
troelsfr: Thanks for picking up on this. I forgot about this one, but obviously it should not be commented out.

WORKSPACE
troelsfr: I am not sure that I understand what you mean by opt-in/opt-out. Adding this to the workspace is the root cause why Windows builds fails (even if the dependency is never used). Once we have a version of this that can be added without failure all other can be uncommented and the docker target will just not be available on Windows but all others will. Hope that makes sense?

swernli: Ah, that does clarify, thanks. I didn't realize this was related to issues of Windows support!

qir/qat/BUILD
swernli: Makes sense, thanks for the background!

PR #153, opened 2022-12-02
Fix handling of `--allow-any-qis` param

swernli:
This channge wires up the internal handling of the `--allow-any-qis` parameter, which was present but not correctly checked before. Now, when the flag is provided, it will override the settings from the loaded adaptor profile and skip checking the qis whitelist, allowing for easier experimentation with new instructions before formally added them to an adaptor profile.

### DIFF
--- a/qir/qat/Passes/ValidationPass/ValidationPass.cpp
+++ b/qir/qat/Passes/ValidationPass/ValidationPass.cpp
@@ -303,7 +303,7 @@ bool ValidationPass::satisfyingExternalCallRequirements()
         for (auto const& k : external_calls_)
         {
             if (allowed_functions.find(k.first) == allowed_functions.end() &&
-                allowed_qis.find(k.first) == allowed_qis.end() && 
+                allowed_qis.find(k.first) == allowed_qis.end() &&
                 !(k.first.rfind("__quantum__qis__", 0) == 0 && qis_config_.allowAnyQis()))
             {
                 logger_->setLlvmHint("");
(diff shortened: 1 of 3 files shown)


qir/qat/Passes/ValidationPass/ValidationPass.cpp
```diff
@@ -295,7 +295,7 @@ bool ValidationPass::satisfyingExternalCallRequirements()
 {
     auto ret = true;
 
-    if (profile_config_.allowlistExternalCalls())
+    if (profile_config_.allowlistExternalCalls() && !qis_config_.allowAnyQis())
```
cesarzc, line 298: I believe this code does not really distinguish between generic external calls and QIS calls.

Perhaps having two distinct `if` statements, one using `allowlistExternalCalls` as conditional which internally checks against `allowedExternalCallNames` for non-QIS functions, and another using `!allowAnyQis` as conditional which internally checks against `allowedQis` for QIS functions. @swernli what do you think?

swernli: I think you make a great point. I adjusted the check to instead only allow functions that start with `__quantum__qis__` when the option is present and verified that other functions still get rejected. This makes the behavior line up with the parameter name more clearly.

cesarzc (approved):

qir/qat/Passes/ValidationPass/ValidationPass.cpp
cesarzc: Thanks for this fix!

PR #154, opened 2022-12-02
Trying smaller devcontainer

swernli:
This updates the devcontainer to use `ubuntu:focal` as the base image, which requires installing a few extra tools but overall is a much smaller resulting image (~1 GB vs ~11 GB with the previous configuration). This makes it a smaller download and lower footprint for both codespaces and local development, with the tradeoff being less software and tooling installed by default.


swernli:
superseded by https://github.com/qir-alliance/qat/pull/156

PR #156, opened 2022-12-14
Move to LLVM 14

idavis:


swernli (commented):
Might be good to update the devcontainer too, even though that needs other work/updates besides the LLVM version.

.github/workflows/trusted-cd-mac-release.yml
```diff
@@ -17,14 +17,15 @@ jobs:
       - name: "Installing dependencies"
         run: |
           rm '/usr/local/bin/2to3'
-          brew install llvm@13
+          brew uninstall llvm@13
```
swernli, line 20: This one has an explicit uninstall of LLVM 13, but the ci-mac-on-main-pr.yml just installs 14. Is it necessary to have this uninstall here, and if so, why isn't it needed for the CI path?

swernli (approved):

qir/qat/Passes/StaticResourceComponent/AllocationAnalysisPass.cpp
```diff
@@ -29,7 +29,7 @@ bool AllocationAnalysisPass::extractResourceId(llvm::Value* value, uint64_t& ret
             return false;
         }
 
-        llvm::Type* element_type = pointer_type->getElementType();
```
billti, line 32: Do we still need lines 26 to 30 above? There's no other use of `pointer_type` with this change. Or do you still need the null check on line 27 (i.e. you can't just check if element_type is null here)

billti (approved):

qir/qat/Passes/StaticResourceComponent/AllocationAnalysisPass.cpp
idavis: @billti FYI, we eventually have to completely change this code for opaque pointers.

PR #162, opened 2023-02-07
Fix reindex after reset, defer measure ordering

swernli:
This change fixes the behavior of the reindex after reset pass, which when run after the defer measurements pass can cause incorrect qubit measurement. Instead of deferring measurement first, which also moves reset calls, the pass ordering now does reset elimination and reindexing first in the default and configured adaptor pipelines.

This also fixes a bug where configuration or cmdline values for adaptor pipeline were incorrectly appended rather than replacing the default values.

### DIFF
--- a/qir/qat/Passes/StaticResourceComponent/ReplaceQubitOnResetPass.cpp
+++ b/qir/qat/Passes/StaticResourceComponent/ReplaceQubitOnResetPass.cpp
@@ -198,12 +198,6 @@ llvm::PreservedAnalyses ReplaceQubitOnResetPass::run(llvm::Function& function, l
                     new_instr = new llvm::IntToPtrInst(new_index, pointer_type);
                     builder.Insert(new_instr);
 
-                    auto op_as_instr = llvm::dyn_cast<llvm::Instruction>(op);
-                    if (op_as_instr)
-                    {
-                        to_remove.push_back(op_as_instr);
-                    }
-
                     instr.setOperand(static_cast<uint32_t>(i), new_instr);
                     already_replaced.insert(new_instr);
                 }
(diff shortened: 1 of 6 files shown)


billti:
Sounds like a few bugs were found/fixed. No tests?

idavis (commented):
@swernli, what was the side effect of this part?

qir/qat/Passes/StaticResourceComponent/ReplaceQubitOnResetPass.cpp
```diff
@@ -198,12 +198,6 @@ llvm::PreservedAnalyses ReplaceQubitOnResetPass::run(llvm::Function& function, l
                     new_instr = new llvm::IntToPtrInst(new_index, pointer_type);
                     builder.Insert(new_instr);
 
-                    auto op_as_instr = llvm::dyn_cast<llvm::Instruction>(op);
-                    if (op_as_instr)
-                    {
-                        to_remove.push_back(op_as_instr);
-                    }
-
```
swernli, line 206: @idavis clarified he was asking about this part of the code, so I'll explain more the motivation here...

This code added instances of the out-of-date LLVM instruction to the list of instructions to remove at the end of the pass. In that if-statement below, the instructions are only removed if they have no uses remaining, which the replacement here ensures. However, there is another bug somewhere in here that causes the same instruction to get added into the list more than once, which will cause a crash when the already removed instruction is removed a second time. As it turns out, these leftover instructions don't need to be pruned here because we run the dead code elimination pass after replace-on-reset, so it was simpler to delete the extraneous removal tracking.

idavis (approved):

PR #166, opened 2023-07-03
Fix outline layout in docs and typos in docs

qartik:
Before:

Screenshot 2023-07-03 at 11 26 57 AM

After:

Screenshot 2023-07-03 at 11 27 21 AM


qartik:
I have no idea about the check that's failing.

qartik:
@idavis / @swernli could you please review this PR?

swernli:
@qartik I think we have the fix ready for the pipeline. We've made the linting check optional and once #167 merges the Q# tests should be passing. I'll file an issue about improving linting and finding another tool besides trunk; in the meantime we'll treat that check as informational only.

swernli (approved):

